PR #14, opened 2017-02-06
Match subsequence instead of entire input sequence.

### LINKED ISSUE #6: a+ pattern in pattern validator doesn't work
The following test case fails as Java regular express returns false. Need a workaround. 
    {
        "description": "pattern is not anchored",
        "schema": {"pattern": "a+"},
        "tests": [
            {
                "description": "matches a substring",
                "data": "xxaayy",
                "valid": true
            }
        ]
    }

mspiegel:
Closes #6. There is the larger issue that json schema regular expressions need to use ECMA 262 regular expressions instead of Java regular expressions. I can open a separate ticket for that. I think other people are solving it by using the Rhino JS engine for Java 7 or the Nashorn JS engine for Java 8.


stevehu:
Cool!!!!! There is always something to learn:) Thanks a lot for your help.

PR #16, opened 2017-02-10
Switch to targeting Java 7

abayer:
As silly as this may sound, I've still got to compile against Java 7 for Jenkins plugins, at least for a while longer. I'd *really* like to use this library in a key Jenkins plugin, but can't unless it's compiled against Java 7. I tried this locally and everything worked fine. Thanks!

(Oh, and I also bumped the version to 0.1.5-SNAPSHOT)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.networknt</groupId>
     <artifactId>json-schema-validator</artifactId>
-    <version>0.1.4</version>
+    <version>0.1.5-SNAPSHOT</version>
     <description>A json schema validator that supports draft v4</description>
     <url>https://github.com/networknt/json-schema-validator</url>
     <name>JsonSchemaValidator</name>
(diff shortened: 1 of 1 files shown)


stevehu:
I have the same problem with swagger-codegen. I couldn't convince their developers that Java 8 can compile Java 7 code:)

I have merged and reverted for this pull request. Is it possible that we create another branch to support Java 7? They is performance difference between Java 7 byte code and Java 8 byte code as this library is used in my light-java microservices framework which I have spent a lot of time to reduce latency. 

I have reviewed your projects and I am amazed with your work and contributions to the community and I want you to use the library. I am just trying to find an alternative solution instead of downgrading the main library to Java 7. 

I am thinking to create another branch and release a library in Java 7 as below. Let me know if this works for you and our last resort will be to downgrade to Java 7. 

<dependency>
    <groupId>com.networknt</groupId>
    <artifactId>json-schema-validator-7</artifactId>
    <version>0.1.4</version>
</dependency>

PR #20, opened 2017-03-28
default messages on empty messages in ValidatorTypeCode

eskabetxe:
on ValidatorTypeCode are some MessageFormat with empty formats that cause a empty message on ValidationMessage.


stevehu:
@eskabetxe Thanks for your help.

PR #22, opened 2017-03-30
fix #21 Use of "id" field does not work for field reference

### LINKED ISSUE #21: Use of "id" field does not work for field reference
I am using a schema to define all the common fields of my objects.
But when I refer to them from some field, the validation does not work correctly, even if it puts erroneous data always comes out that is valid.

subschema with fields defenition
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "someSubschemaField": {
          "type": "integer",
           "minimum": 1,
          "maximum": 99999
   }
}

schema that uses subschema
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "https://www.example.com/subschema.json",
  "type": "object",
  "properties": {
    "someSchemaField": {
       "$ref": "#/someSubschemaField"
    }
}

using http://www.jsonschemavalidator.net/
validates the schema correctly

eskabetxe:


stevehu:
Thanks @eskabetxe for your contribution. There are some existing test cases failed with the change and I am planning to look into it once I get time. It would be great if you could take a look what the root cause and fix it. Happy coding!

eskabetxe:
@stevehu resolved, sorry i take a wile to understand the problem.. 
there are some issues if the "id" as the same value as "schema" causing some infinite loop

stevehu:
@eskabetxe Thanks a lot for your help.

stevehu:
@eskabetxe I am wondering if you could help to add a test case for the subschema to ensure other people's changes won't break your code.

eskabetxe:
I had already thought about it. 
The problem is that it will be necessary a url with a valid scheme
Could we use a url to a file in repository (github)?

stevehu:
Not necessary. I have undertow included in the test scope already and going to start a test server for this purpose. I have some other remote schemas that need to be tested as well. Give me some time and I will put it in. Hopefully this weekend. Thanks for reminding me:)

stevehu:
@eskabetxe As per our discussion, I have added undertow server in the test for remote references. Let me know if you have any questions. Thanks.

eskabetxe:
@stevehu sorry for the delay..
I didnt have time available before, i will check this asap

PR #30, opened 2017-04-26
Features/one of validator

eskabetxe:
f schema not valid to oneOf, added all error, that isnt ADDITIONAL_PROPERTIES


stevehu:
@eskabetxe Thanks a lot. I will test it and release it soon.

stevehu:
These edits and crossedits are not been used as far as I know. I added them to extend json schema validation but later removed it in order to make sure that the implementation is the same with others.

stevehu:
Managed to release 0.1.7 tonight. Thanks a lot for your help. 

https://github.com/networknt/json-schema-validator/releases/tag/0.1.7

eskabetxe:
thanks @stevehu

stevehu:
@bamateus Thanks for bringing this up. I will add your test case and debug into it. If this is the root cause, it should be an easy fix.

eskabetxe:
hi @bamateus 
the "additionalproperties" are remove, if none of the schemas are validated..

this is because, when you have "oneof" with more than one schema, and the schema is not valid for none of the definitions,  will result in to many errors of "additionalproperties".

i see that in your example there are only one schema in oneOf, and i undertood that this behavior could be wrong..

we could check that if only exists one schema add all errors, or remove that condition...

bamateus:
Hello  @eskabetxe and @stevehu, thank you for your answer.

My test case was just an example, on our real scenario we have 12 schemas in oneOf. Currently, we are sending to our Client integration team, the validator returned messages. We want to send all the information that is wrong at once.

In my opinion, if there are no limitations like technical/performance issues, I think the validator should not "hide" any information. In this case, our schemas have `"additionalProperties" : false` that's why we were expecting to see all the errors. I don't know if the JSON Schema draft-04 says anything about this.

If you don't agree or think it's better to keep it as is, maybe we can have some kind of property to activate/deactivate this behaviour.

Thanks

eskabetxe:
In my case we have 8 schemas with "additionalProperties" false..
the 8 schemas are totally diferents, and in my case if the json is not valid in any schema, this errors simply dont tell anything.

as example, if we have 6 schemas with 20 fields diferents, we wil have aprox 100 messages of "additionalproperties" errors..

if this is the expect behavior, for me is ok as the erros can be filter later, and its a easy fix (simple remove de condiction)..

@stevehu what you think?

stevehu:
This behaviour is not defined in the draft v4 as there is no test case from jsonschema.org. It is purely implementation details. I think both of your expected behaviours are valid given different use cases. Just like #33, I will introduce a config file for the validators to customize it according to different use cases. There are some other place like Java expr vs Javascript expr as well. I will refactor the library to use SPI to load validators so that people can add their own validator if they want.

PR #37, opened 2017-06-26
Issue 26

eskabetxe:
adding relative $ref urls


eskabetxe:
@stevehu could you release this?

src/main/java/com/networknt/schema/RefValidator.java
```diff
@@ -66,16 +63,47 @@ public RefValidator(String schemaPath, JsonNode schemaNode, JsonSchema parentSch
...
+        } else {
+            JsonNode node = parentSchema.getRefSchemaNode(refValue);
+            if (node != null) {
+                schema = new JsonSchema(mapper, refValue, node, parentSchema);
+            }
+        }
+    }
+    
+    private boolean isRelativePath(String schemaUrl) {
+    	return !schemaUrl.startsWith("http");
```
whiskeysierra: What about relative **local** paths? With this change I can't load local schemas anymore using `$ref`.

src/main/java/com/networknt/schema/RefValidator.java
```diff
@@ -66,16 +63,47 @@ public RefValidator(String schemaPath, JsonNode schemaNode, JsonSchema parentSch
...
+    }
+    
+    private String obtainAbsolutePath(JsonSchema parentSchema, String schemaUrl) {
+    	String baseSchemaUrl = parentSchema.findAncestor().getSchemaNode().get("id").textValue();
```
whiskeysierra: Where does this `"id"` property come from and why is it treated as a URL?

whiskeysierra: I found https://spacetelescope.github.io/understanding-json-schema/structuring.html?highlight=dependencies#the-id-property. But it should be optional, I'd say. I still want to use $ref on local relative files.

stevehu:
@whiskeysierra I hear you. There are too many version of specs and it is very hard to support all of them at the same time with the current architecture. I am planning to refactor it with service from light-4j so that different user can choose their implementations. 

https://github.com/networknt/json-schema-validator/issues/54

PR #38, opened 2017-07-20
Don't leave out 50% of your audience who still use older java

basinilya:
Only few lines in OneOfValidator.java use constructs specific to java 1.7 and 1.8. I rewrote them with older syntax.

Unfortunately, the dependencies of junit tests of this project are compiled for java 1.7, so for the tests to work out of the box in Eclipse I had to mention both versions in maven-compiler-plugin config. Eclipse picks the latest one for the project execution env.

We, personally, use java 1.7 in our project and we need this library.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,8 @@
     </distributionManagement>
 
     <properties>
-        <java.version>1.8</java.version>
+        <java.version>1.6</java.version>
+        <java.testversion>1.7</java.testversion>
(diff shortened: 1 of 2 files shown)


stevehu:
@basinilya Thanks a lot for your effort. I will test it with my frameworks that are depending on this lib and release a new version.

stevehu:
I have released a new version and the release note can be found at https://github.com/networknt/json-schema-validator/releases/tag/0.1.10. Thanks a lot for your help.

PR #43, opened 2017-10-13
Replaced 'new URL' with a call to a new URLFactory 

kenwa:
The original implementation supports schema's to be accessible either in http, http, file, etc.. This pull request adds the possibility to also refer to schema's present in the applications classpath.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.networknt</groupId>
     <artifactId>json-schema-validator</artifactId>
-    <version>0.1.10</version>
+    <version>0.1.10-escenic</version>
     <description>A json schema validator that supports draft v4</description>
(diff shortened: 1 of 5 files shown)


stevehu:
@kenwa Thanks for your help. It is a very nice feature to support loading schema from classpath.

kenwa:
Nice that you liked the feature :-) Just wanted to let you know that a bump of version number accidentally slipped into the PR. You should probably fix that one. 

Any chance that you could create a new release of the library? That would make it possible for us to use that one instead of our forked variant.

stevehu:
@kenwa I have released 0.1.11 minutes ago. I think this is an very unique feature and we need to let other people to enjoy it. Could you please write a unit test case to show people how to use it? The test case can also prevent other PRs to break this feature in the future. Thanks a lot for your contribution.

kenwa:
@stevehu Great! Then we'll use that one instead :-) 

I created a PR with a test case showing how to use the feature a few minutes ago.

PR #45, opened 2017-10-23
Added test case for loading schemas from classpath

kenwa:


kenwa:
No problem. Just happy to help.

PR #47, opened 2017-11-13
Change access modifiers in ValidationMessage

ehrmann:
Make getArguments public and setArguments package-private.


stevehu:
@ehrmann Thanks a lot for your help. I will change other setters and getters in the class in case someone is accessing it from other packages.

PR #49, opened 2017-11-14
Use LinkedHashSets for ValidationMessages

ehrmann:
For things that might be read by humans, maintaining order can be
valuable. This change uses LinkedHashSet rather than HashSet to
maintain the iteration order of error in the order they're encountered.

Additionally, there are a few optimzations in validators:
- Using singletonSet and emptySet rather than a *HashSet when
  appropriate
- Map iteration optimization
- StringBuilder optimization


stevehu:
@ehrmann Thanks a lot for you help. I went back to the collections again and learned more this time. There is always something to learn from the community and I love it. I will test it with some of the applications and release a new version soon.

PR #50, opened 2017-11-22
Support custom meta schemas with custom keywords and formats

kkalass:
Thanks for this nice library!

I want to switch a project from fge where I used the extensibility of fge extensively. So I refactored your library a bit to allow for multiple meta schemas, custom keywords etc. 

I would be happy if you would accept my changes.

Refactored so that the library supports multiple meta schemas and users
of the library can add custom meta schemas as well as controlling
aspects like e.g. fetching via a URL.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,9 +13,9 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-
-<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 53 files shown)


stevehu:
@kkalass Thanks a lot for your contribution. We have other projects that need to be released due to customer's demands and will review/test the changes tomorrow. A new release should be out during this weekend.

jocnud:
@kkalass  Is there any doc which we can use to create custom validator. I am using this api, and in a want to write a custom validator

kkalass:
Hi @jocnud I just added the PR https://github.com/networknt/json-schema-validator/pull/89 which includes at least a test case that shows how to implement a custom validator. I hope that helps?

PR #52, opened 2017-11-27
References that cannot be resolved should be treated as an error

kkalass:
Documents were validated even if "$ref" pointed to a non-existing schema. 

The json schema spec does not explicitely  state how to treat this case, but I would like to argue that a non-existing referenced schema should be treated as an error.


kkalass:
Please note that the "optimization of oneOf" PR contains this change as well. So if you agree to both PRs, you do not need to merge this one but the "optimization of oneOf" only.

PR #53, opened 2017-11-27
Optimization for OneOf

kkalass:
An often-used pattern in conjunction with oneOf is to use a single value
enum as a constant where each oneOf alternative typically has a different
constant.

If the corresponding property is set in the json node (and thus is
textual) we can skip all alternatives that have a different constant
value defined for that property since they will definitely not match.

For schemas with many alternatives and a lot of items that need to run
through those alternatives, I found this to speedup validation
considerably (30% faster in my case).

Please note that this commit contains the semantics from PR "unresolved-refs-are-errors" as well - if you object to that one, I can adjust this PR.


stevehu:
@kkalass Thanks a lot for your help. The update is pretty significant and I have to review it multiple times to ensure no API is broken. Also, the last several weeks were crazy for me with other projects. I will do some future tests and release it.

PR #57, opened 2018-01-09
Optimize AdditionalPropertiesValidator

ehrmann:
This change makes fields final and changes `allowedProperties` to a `HashSet` so lookups are in O(1) time.


stevehu:
@ehrmann Thanks a lot for your help. The changes make perfect sense for these two validators.

PR #61, opened 2018-02-13
Restore validator type code from value

ehrmann:
79acd4d0 removed the fromValue method from ValidatorTypeCode. Restore it to its state before 79acd4d0.


stevehu:
@ehrmann Thanks a lot for your help. That particular commit was massive and I missed these two static methods while doing code review. How did you find it out?

ehrmann:
I was using it :)

stevehu:
Thanks God!!! Could you please add a test case so that nobody can remove it accidentally? Thanks.

PR #66, opened 2018-02-17
Only apply pattern validation on strings

ehrmann:
Currently, `pattern` is applied to strings, numbers, and integers. According to the [spec](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.3.3),

> A string instance is considered valid if the regular expression matches the instance successfully. 

I think that means patterns are ignored for non-strings.


stevehu:
That makes sense after re-read the spec. However, I didn't cross check the older version of spec. Thanks a lot for your help.

PR #69, opened 2018-03-03
Only apply required validator to objects

ehrmann:
The spec for `require` say

>  An object instance is valid against this keyword if every item in the array is the name of a property in the instance.

`require` shouldn't be applied to non-objects.


stevehu:
@ehrmann Thanks a lot for your help. I will test it and release the new version asap.

PR #70, opened 2018-03-03
Minor optimizations

ehrmann:


stevehu:
@ehrmann 0.1.16 is released and will be available from Maven Central shortly. Thanks.

PR #76, opened 2018-03-15
Make remaining JsonSchema constructors public

kkalass:
To improve extensibility by library users, it would be very helpful if one can use the full JsonSchema constructor and thus create a JsonSchema for a part of a schema and explicitely validate only that part.


stevehu:
@kkalass Thanks a lot for your help. It makes sense now. In the beginning, I was thinking to use factory instead of construct JsonSchema directly. Since we have a public constructor already, it makes sense to make all of them public. The end user has a choice to use constructor or factory/builder.

PR #77, opened 2018-03-15
One of was broken - it did not fail when there were no valid schemas

kkalass:
There was actually a bug in my previous commit for shortcut validation of oneOf - we reported success even if none of the schemas validated.  Sorry for sending you a patch without appropriate test case.

Here is the test case and a fix for that one.

But I noticed that https://github.com/networknt/json-schema-validator/commit/a77673858c7df3f2b08eedb6b971cd74fda58920#comments seems to have originally introduced the behaviour I called "bug" above deliberately. Unfortunately the question posted by @bamateus underneath that commit was not answered. As far as I understand the json schema spec, we must not validate oneOfs when none of the schemas validates. And if the validation error is about an additional property, it still is a validation error (if it was reported wrongly then it should be fixed in the validation code for additional properties, not here).

### DIFF
--- a/src/main/java/com/networknt/schema/OneOfValidator.java
+++ b/src/main/java/com/networknt/schema/OneOfValidator.java
@@ -141,14 +141,14 @@ public class OneOfValidator extends BaseJsonValidator implements JsonValidator {
                 continue;
             }
             JsonSchema schema = validator.schema;
-        	Set<ValidationMessage> schemaErrors = schema.validate(node, rootNode, at);
+        	    Set<ValidationMessage> schemaErrors = schema.validate(node, rootNode, at);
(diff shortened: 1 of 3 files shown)


stevehu:
@kkalass Thanks a lot for the patch. oneOf is not a widely used validator and that is why we didn't find the issue. It is really good to add test cases for every PR and we should encourage other developers to do the same.

kkalass:
great, thanks for accepting my patch(es)!

2018-04-04 21:56 GMT+02:00 Steve Hu <notifications@github.com>:

stevehu:
@kkalass We are all working together to make open source projects more useful for the community.

PR #106, opened 2018-12-11
Fix for https://github.com/networknt/light-rest-4j/issues/64

BalloonWen:
-this PR is part of the fix for https://github.com/networknt/light-rest-4j/issues/64
-BaseValidator will have a new property "ValidatorConfig", this ValidatorConfig will be set when constructing from the constructor, get from ValidationContext. Its child validator "TypeValidator" will use this property to get the config.
-"VaidatorConfig" will be set into "ValidationContext" when creating new JsonSchema using "newJsonSchema()" in "JsonSchemaFactory"

### DIFF
--- a/src/main/java/com/networknt/schema/BaseJsonValidator.java
+++ b/src/main/java/com/networknt/schema/BaseJsonValidator.java
@@ -18,13 +18,12 @@ package com.networknt.schema;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.networknt.schema.url.URLFactory;
+import com.networknt.schema.url.ValidatorConfig;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 
 import java.net.MalformedURLException;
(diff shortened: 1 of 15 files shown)


BalloonWen:
renamed ValidatorConfig to SchemaValidatorsConfig based on comments of https://github.com/networknt/light-rest-4j/pull/65

stevehu:
Any reason to use plural instead of singular? SchemaValidatorsConfig vs SchemaValidatorConfig.

BalloonWen:
> Any reason to use plural instead of singular? SchemaValidatorsConfig vs SchemaValidatorConfig.

Because this config obj is in BaseJsonValidator. Any validator extends this BaseJsonValidator will have this config, so I think this config obj can contain configs not only for TypeValidator but can also have configs for other validators. What do you think?

stevehu:
Good. It is just a name and I am just curious. Thanks.

PR #105, opened 2018-10-31
temporary fix to performance issue

nitin456:

PR #107, opened 2018-12-11
Fix for perfomance issue

nitin456:
We observed that there is significant drop in performance in due to unnecessary validate call to anyOf schemas. The fix is to avoid validate if there is mismatch in the type.  This PR has fix along with test case.


nitin456:
@stevehu please review and merge this PR. This PR is continuation of earlier PR https://github.com/networknt/json-schema-validator/pull/105

I have also found that there is some performance drop due to creation of too many LinkedHashSet Objects. It can be observed on the large JSON structure.

nitin456:
@stevehu I have fixed https://github.com/networknt/json-schema-validator/issues/108 in this PR. I have also added test cases around union type. Please review it.

PR #112, opened 2019-01-07
AnyOfValidator: only return expectedTypeList if not empty

c14s:


nitin456:
@c14s, you are correct. It should be expectedTypeList. Thanks!

stevehu:
@c14s @nitin456 @ddobrin Thanks lot for your help.

PR #113, opened 2019-01-17
Fixed validation for path parameters and query parameters

jiachen1120:
@NicholasAzar @stevehu Since the RequestValidator is built on json-schema-validator. There is another PR related to this issue can be found in networknt/light-rest-4j repo. These changes in different repo contributions to the same issue.

RFC which specifies the details of fix can be found here: https://github.com/networknt/light-rfcs/blob/master/light-rest-4j/0001-openapi-validator-bug-fix.md

### DIFF
--- a/src/main/java/com/networknt/schema/EnumValidator.java
+++ b/src/main/java/com/networknt/schema/EnumValidator.java
@@ -65,11 +65,24 @@ public class EnumValidator extends BaseJsonValidator implements JsonValidator {
 
         Set<ValidationMessage> errors = new LinkedHashSet<ValidationMessage>();
 
-        if (!nodes.contains(node)) {
+        if (!nodes.contains(node) && !isTypeLooseEqual(node)) {
             errors.add(buildValidationMessage(at, error));
         }
 
         return Collections.unmodifiableSet(errors);
+    }
+
+    private boolean isTypeLooseEqual(JsonNode node) {
+        if (config.isTypeLoose() && TypeFactory.getValueNodeType(node) == JsonType.STRING) {
+            String nodeText = node.textValue();
+            for (JsonNode n : nodes) {
+                String value = n.asText();
+                if (value != null && value.equals(nodeText)) {
+                    return true;
+                }
+            }
+        }
+        return false;
     }
 
 }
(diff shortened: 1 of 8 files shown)


src/main/java/com/networknt/schema/EnumValidator.java
```diff
@@ -1,75 +1,88 @@
...
+        Set<ValidationMessage> errors = new LinkedHashSet<ValidationMessage>();
+
+        if (!nodes.contains(node) && !isTypeLooseEqual(node)) {
+            errors.add(buildValidationMessage(at, error));
+        }
+
+        return Collections.unmodifiableSet(errors);
+    }
+
+    private boolean isTypeLooseEqual(JsonNode node) {
```
NicholasAzar: 1. I think it's doing a lot more then the name would suggest; It's checking if the type is loose, checking if the parameter is a string node, and checking if that string value appears in the nodes instance variable (the enum).. similar to an "contains" check. Maybe we can rename it to something like "isTypeLooseAndInEnum"? or move the typeLoose check out of it and just have this method check if the node is a string and in the enum?
2. Would it make sense to some unit tests to it?

jiachen1120: Yes, this name does not quite match its function. I will modify it. I feel that taken out typeLoose check separately to be more understandable and more consistent with the judgments in other verifications.

I will try to add some testcase about enum
For example, enter a number, enter a string, and include a number or string in the enum. Test if it works in the case of typeLoose.

src/main/java/com/networknt/schema/MaximumValidator.java
```diff
@@ -1,70 +1,72 @@
...
+
+        parseErrorCode(getValidatorType().getErrorCodeKey());
+    }
+
+    public Set<ValidationMessage> validate(JsonNode node, JsonNode rootNode, String at) {
+        debug(logger, node, rootNode, at);
+
+        if (!node.isNumber() && !(config.isTypeLoose()
+                && TypeFactory.getValueNodeType(node) == JsonType.STRING
+                && TypeValidator.isNumeric(node.textValue()))) {
```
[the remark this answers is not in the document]
jiachen1120: Ok, I think to integrate typeLoose type validation into the type validator may make sense?

NicholasAzar:
@stevehu This is a dependency to the linked PR

PR #119, opened 2019-02-21
Almost JSON-spec compliant validation of numeric values

### LINKED ISSUE #118: Support numerals the way JSON spec defines them
Currently [`TypeValidator`](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/TypeValidator.java#L65) has a loose mode (which is great) but [`isInteger`](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/TypeValidator.java#L101) and [`isNumeric`](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/TypeValidator.java#L125) do not fully comply with numeric literals defined by [JSON spec](http://json.org/). In particular the following values are not regarded as numbers by JSON (some of them are successfully validated by `TypeValidator`)

```
"01.1", "1.", ".1", "E1", "E+1", "E-1", ".E1", ".E+1", ".E-1", ".1E1", ".1E+1", ".1E-1", "+1"
```

also  the following values are all valid numbers as per spec

```
"1", "-1", "1.1", "-1.1", "0E+1", "0E-1", "0E1", "-0E+1", "-0E-1", "-0E1", "0.1E+1", "0.1E-1", "0.1E1", "-0.1E+1", "-0.1E-1", "-0.1E1", "10.1", "-10.1", "10E+1", "10E-1", "10E1", "-10E+1", "-10E-1", "-10E1", "10.1E+1", "10.1E-1", "10.1E1", "-10.1E+1", "-10.1E-1", "-10.1E1" 
```

**Note**
`+1`, `0001` is currently successfully validated by `TypeValidator` adhering to spec might introduce breaking change for some of existing users

kosty:
Fixes #118 , yet leaves `+1` as a valid option

### DIFF
--- a/src/main/java/com/networknt/schema/TypeValidator.java
+++ b/src/main/java/com/networknt/schema/TypeValidator.java
@@ -102,6 +102,9 @@ public class TypeValidator extends BaseJsonValidator implements JsonValidator {
         if (str == null || str.equals("")) {
             return false;
         }
+
+        // all code below could be replaced with
+        //return str.matrch("[-+]?(?:0|[1-9]\\d*)")
         int i = 0;
         if (str.charAt(0) == '-' || str.charAt(0) == '+') {
             if (str.length() == 1) {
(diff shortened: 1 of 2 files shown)


stevehu:
[Almost JSON-spec compliant validation of numeric values](https://trello.com/c/rDKxyu8r/105-almost-json-spec-compliant-validation-of-numeric-values)

kosty:
@jiachen1120 Thanks for feedback. To provide some more context I tried to be as close to original JSON spec which actually points to those as invalid numeric representations. 

With that said, I am happy to have numeric parsing in any consistent way. Would `BigDecimal` be consistent with `go`, `js`, `python` et. al. environments?
Also, in that case should we stick to `BigInteger` as validation for integers?

stevehu:
@kosty @jiachen1120 I have learned a long time ago that it is very hard to ensure that the JSON schema validation works exactly the same for all languages. Different languages have different limitations and sometimes we have to accept the differences. There are still several open issues in Java that we cannot match the Javascript implementation, but these are all edge cases and we never bothered. 

For me, I am trying to avoid using try/catch as it has performance issue compare with other approaches whenever necessary. For this particular case, I think we are OK if we can cover all possible cases without using try/catch. @NicholasAzar @ddobrin What do you think?

kosty:
@stevehu @jiachen1120 I do agree on performance hit with try/catch. Arguably pattern matching has a similar downside. Also good point on challenges achieving consistency across different languages.

ddobrin:
I am generally avoiding this try-catch approach and prefer the "exception barrier" concept when an exception is left to bubble up.

Otherwise we pay the price for each call, regardless of pass/fail.

NumberFormatException is a RuntimeException and it would be caught if that where to happen, at the latest by the ExceptionHandler in the exception module

stevehu:
The exception will be caught in the same block, so it won't bubble up. I am wondering how other languages handlers this kind of problem if they don't use the try/catch block. I'd prefer a generic solution; however, I am not against the try/catch if it ends up the simplest solution.

ddobrin:
I was suggesting to let it up bubble up :) and maybe go that route.

kosty:
Did some googling on try/catch performance and it seems to be the `throw` and all the stack-walking to blame:

[stackoverflow answer claiming ~66 times slowdown](https://stackoverflow.com/a/299315)
[more stackoverflow](https://stackoverflow.com/a/16451908)
[a blog post with benchmarks and graphs](https://mattwarren.org/2016/12/20/Why-Exceptions-should-be-Exceptional/)

In my case validator is quite deep in the stack, thus exception might be really harming overall performance. Also result seems predictable, so exception feels more like an overkill

jiachen1120:
@kosty Thank you for sharing. As summarized below, 

the Java virtual machine uses the method call stack to keep track of a series of method invocation procedures in each thread. The stack holds local information for each method called. Each thread has a separate method call stack. For the main thread of the Java application, the bottom of the stack is the main() method of the program's entry method. When a new method is called, the Java virtual machine puts the stack structure describing the method on top of the stack, and the method at the top of the stack is the method being executed.

When a method completes normally, the Java virtual machine pops the stack structure of the method from the call stack and continues processing the previous method. If an exception is thrown during the execution of a method, the Java virtual machine must find a catch block that catches the exception. It first looks to see if the current method has such a block of code. If it exists, then the block is executed. Otherwise, the Java virtual machine pops the stack structure of the method from the call stack and proceeds to the previous method to find the appropriate capture code block. In the backtracking process, if the Java virtual machine finds a block of code that handles the exception in a method, the stack structure of the method becomes the top element of the stack, and the program flow goes to the exception handling code part of the method to continue execution.

If the code block that throws the exception and the code block that catches the exception are in the same place, the impression will be smaller; if the Java virtual machine must search the method call stack to find the exception handling code block, the impact on performance will be greater. Especially when the exception handling code is at the bottom of the call stack, the Java virtual machine locating exception handling code requires a lot of work.

Therefore, maybe we should not use the exception handling mechanism to control the normal flow of the program when the exception is not handled in place. 

Thanks again for sharing, let me know more about the exception handling. 👍

stevehu:
@kosty Thanks a lot for your help and I have learned a lot from the discussion.

stevehu:
@kosty @jiachen1120 @ddobrin I have released 1.0.4 to the maven central and will update the light-4j framework to depend on this version in the next release. Thanks a lot for your help. 

https://github.com/networknt/json-schema-validator/releases

Issue #124, opened 2019-03-21
Alternate source for remote schema


rhwood:
I have a need to be able to validate JSON schemas using an isolated system that cannot reach the internet. Since all schemas I need to validate include `"$schema": "http://json-schema.org/draft-04/schema#"`, this is causing a java.net.ConnectionException. Is there a way to map URLs to a local resource, so that we could have a local copy of an externally defined schema (similar to the [CatalogResolver](https://docs.oracle.com/javase/9/docs/api/javax/xml/catalog/CatalogResolver.html) for XML processing in the JRE)?

stevehu:
If your local resource is a valid URL, then it is OK. JSON schema doesn't support local file system and we have to implement something like an extension to the spec. I remember somebody might have explored it already and gave up.

rhwood:
Accidentally closed.

rhwood:
Do you mind if I create a PR that allows for this outside the specification?

stevehu:
Yes. Please.

PR #125, opened 2019-03-22
feat: Add URL mappings

rhwood:
URL mappings allows public, normally internet accessible URLs (such as might be externally referenced in a schema) to be mapped to local URLs that might be on the filesystem, at a different external URL, or embedded within a JAR file.

This allows the JsonSchema validator to validate a schema with an external reference to validate against the local copy of that reference when the external reference is not available. Note that when using a mapping, the local copy is always used, and the external reference is not queried.


rhwood:
This includes tests that the URL mappings work, both in terms of being schema valid, and by referencing a non-existent URL.

@stevehu two questions:
- [ ] Where should the schema for a JSON file containing mappings be published? 22efdfe uses `https://raw.githubusercontent.com/networknt/json-schema-validator/master/src/main/resources/url-mapping.schema.json` which, while being valid if this PR is merged from that commit without changes, feels like it should be published elsewhere (or use a different URL even if not published at that URL).
- [ ] Should the mapping file `src/test/resources/tests/url_mapping/url-mapping.json` be included in sources, and not just tests, as a usable mapping file to avoid queries against the Draft 4 schema and the mapping schema?

Fixes #124

stevehu:
@rhwood I like the solution which gives users more flexibility for other schema remote references. Regarding the configuration, we have an external config file and config class already and wondering if you could use that instead. This will give users a consistent way for configuration and allow them to leverage the feature easily. 

https://github.com/rhwood/json-schema-validator/blob/master/src/main/java/com/networknt/schema/SchemaValidatorsConfig.java

The config module we are using is part of the light-4j framework and it is tiny. 

https://github.com/networknt/light-4j/tree/master/config

https://doc.networknt.com/concern/config/

rhwood:
@stevehu 3c27e9b changes this to use a `SchemaValidatorsConfig` to provide the mapping. It also allows the `JsonSchemaFactory.Builder` to provide the mapping, merging the two mappings together if both a Builder and a Config provide the mapping.

I've removed mechanisms other than providing the mapping as a `Map<URL, URL>` to a `SchemaValidatorsConfig` or `JsonSchemaFactory.Builder` from the code.

jiachen1120:
@stevehu Currently, we cannot use the config module inside the Json-schema-validator, since the Json-schema-validator is not depend on light-4j, but light-4j depend on Json-schema-validator. 

Therefore, in order to make this url-mapping configurable, maybe we should load a config called "url-mapping.yml" in `light-rest-4j/openapi-validator` and pass this map into an overload method `JsonSchema.validate(content, url-mapping)`.

Otherwise, we can use @rhwood's method, to parse a Json file of url-mapping (`getUrlMappingsFromUrl()` in test) without using config module. 

What do you think?

jiachen1120:
@rhwood Good morning! I just tested the solution by retrieve schema from public url without url-mapping. Then I got the following exception:

com.networknt.schema.JsonSchemaException: java.io.FileNotFoundException: https://raw.githubusercontent.com/networknt/json-schema-validator/master/src/test/resources/tests/url_mapping/url-mapping.schema.json

	at com.networknt.schema.JsonSchemaFactory.getSchema(JsonSchemaFactory.java:228)
	at com.networknt.schema.JsonSchemaFactory.getSchema(JsonSchemaFactory.java:233)
	at com.networknt.schema.UrlMappingTest.testBuilderUrlMappingUrl(UrlMappingTest.java:40)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.io.FileNotFoundException: https://raw.githubusercontent.com/networknt/json-schema-validator/master/src/test/resources/tests/url_mapping/url-mapping.schema.json
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1890)
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1492)
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:263)
	at java.net.URL.openStream(URL.java:1045)
	at com.networknt.schema.url.StandardURLFetcher.fetch(StandardURLFetcher.java:14)
	at com.networknt.schema.JsonSchemaFactory.getSchema(JsonSchemaFactory.java:211)
	... 24 more

My Internet can work properly, Is there any specify setting should be done or retrieve from public url directly is not support? Look forward to your response. Thanks!

rhwood:
The file https://raw.githubusercontent.com/networknt/json-schema-validator/master/src/test/resources/tests/url_mapping/url-mapping.schema.json does not exist (try getting to it through any browser) and will not exist until this PR is merged, so you should be getting that error unless you use a mapping to an alternate source.

Randall Wood

stevehu:
@jiachen1120 I got your point. For the light-4j frameworks, we can load single or multiple configuration files without any issue with light-4j config module. However, it would be very hard for other users to load config files if they are just using the json-schema-validator. I have reviewed SchemaValidatorsConfig and found that some of the attributes are useful for independent users. I would prefer to have a single config file like @rhwood implemented. This is the second option you have described above. Do you see any problem with this approach? If not, could you please approve the PR and I will merge it.

stevehu:
@rhwood Thanks a lot for your help.

Issue #132, opened 2019-04-08
minimum/maximum validation of integral numbers prone to overflow


kosty:
### Description 
Currently both `MaximumValidator` and `MinimumValidator` both rely on `double` as their base type. This causes some omissions for `integer` numbers, e.g. `9223372036854775807` and `9223372036854775806` are both rounded up to `9223372036854776000` which leads to bug in validation when former is the value validated and latter is the `maximum` threshold in schema. Corresponding issues is demonstrated by `-9223372036854775808` and `-9223372036854775807` for `minimum`.

### Sample test
```
    @Test
    public void testIntegralMaximum() throws IOException {
        ObjectMapper m = new ObjectMapper();

        String schema = "{ \"$schema\":\"http://json-schema.org/draft-04/schema#\", \"type\": \"integer\", \"maximum\": 9223372036854775806 }";
        JsonSchema v = JsonSchemaFactory.getInstance().getSchema(m.readTree(schema));
        JsonNode doc = m.readTree("9223372036854775807");

        Set<ValidationMessage> messages = v.validate(doc);
        assertFalse(messages.isEmpty());
    }
```

### Notes
Since `json-schema-validator` is optimized for speed of validation, conditional statements inside `validate` method are not desirable. Similarly, comparison of primitive numeric types is preferred to `BigInteger` / `BigDecimal`.

stevehu:
@stevehu @jiachen1120 Thanks a lot for the effort. I have released a new version 1.0.6 minutes ago and upgraded light-4j framework with the new version with light-bot. 

https://github.com/networknt/json-schema-validator/releases

PR #133, opened 2019-04-08
maximum/minimum validation of integral numbers

kosty:
Addressing issue #132 

Test coverage added to elaborate on various cases. 

It turns out `jackson` differentiates between `int`, `long` and `BigInteger` out of the box. However difference between `double` and `BigDecimal` is an option turned off by default. Furthermore, when enabled `DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS` still parses value as `double` by default and later converts to `BigDecimal`. Thus values greater then `Double.MAX_VALUE` are interpreted as `Infinity`


stevehu:
[maximum/minimum validation of integral numbers](https://trello.com/c/MVzDhVxm/163-maximum-minimum-validation-of-integral-numbers)

PR #137, opened 2019-04-23
fix: avoid performance penalty of using URL in map or set

### LINKED ISSUE #136: Maps of URLs can have performance impacts
While incorporating the feature of URL mappings into an app, [Spotbugs](https://spotbugs.readthedocs.io/en/stable/) flagged the [inefficient use of URL in maps](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-maps-and-sets-of-urls-can-be-performance-hogs-dmi-collection-of-urls).

While this is an easy fix, it does make a breaking change in the API since 1.0.5.

rhwood:
See https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-maps-and-sets-of-urls-can-be-performance-hogs-dmi-collection-of-urls

Fixes #136 by introducing breaking change to the URL mapping feature of versions 1.0.5 thru 1.0.7

### DIFF
--- a/src/main/java/com/networknt/schema/JsonSchemaFactory.java
+++ b/src/main/java/com/networknt/schema/JsonSchemaFactory.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.networknt.schema.url.StandardURLFetcher;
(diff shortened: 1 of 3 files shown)


stevehu:
@rhwood I have merged the previous PR and found this one has a conflict. Could you please help to resolve the conflict? Thanks.

rhwood:
@stevehu I've resolved the conflict

Issue #98, opened 2018-09-07
RefValidator on non-http (classpath & other custom protocol) URL breaks schema loading


gotchazipc:
Hi. I found RefValidator resolves a bad URL when a $ref value points non-http URL. 

For classpath:/ URL, 
When schema a having id `classpath:/some_path/a.json` has a $ref value to `classpath:/some_path/b.json`, URLFetcher tries to load `classpath:/some_path/classpath:/some_path/b.json` 
it seems to be a bug of `RefValidator#obtainAbsolutePath()` 

Changing `$ref` value to '/some_path/b.json' was looked promising but didn't help. `RefValidator#obtainAbsolutePath()` returned `classpath//some_path/b.json`, lacking colon. After changing $ref value to './b.json', it worked as expected. 

For some custom URL, like some_protocol:/some_path/some_file_name.json, I met similar but more complex problems. After digging into code, I found 

1) As `classpath:/...` URLs, absolute URL is treated as relative. RefValidator regards any URL string which does not start 'http' as an relative URL. 
 
2) URLFactory.toURL() always refers ClasspathURLStreamHandler and there's no way to override the factory. That means, the custom protocol URL string cannot be changed into URL object but throws MalformedURLException. Then, RefValidator catches the exception and it tries to interpret whole URL string as some resource path. Finally some unexpected NPE is thrown in somewhere else. 

First one could be solved easily but second problem looks a little bit serious. Relying on URL & URLConnection to load sub-schema doesn't look so good, for managing stream handler factory and many properties on making/managing connections using URL is static or VM-wide, using system properties. Even though URLFetcher can be customized using interface & JsonSchemaFactory.builder(), it's not possible to load URL with custom scheme. 

To provide more flexible customization on loading schema, I hope we could have better design. URLFetcher need to accept string instead of URL, and internal code may need to be changed using URI instead of URL.

stevehu:
@gotchazipc I agree with you that reference URL needs to be handled correctly. I have done a lot of work in the openapi-parser for parsing and validation. Here is a utility module that I think would help to extend the functionality. Need to find time to enhance this library. 

https://github.com/networknt/light-4j/tree/develop/http-url

stevehu:
This issue should be resolved by PR #141

Issue #86, opened 2018-05-02
RefValidator baseSchemaUrl Port Ignored


jawaff:
Hello,

I'm not sure if this is an issue, but it might be. I just thought I'd bring it up to see if it really is an issue.

Anyway, I was attempting to define an "id" property in a test schema. That defines the base URL that "$ref" properties will build off of. Then I used a "$ref" property and expected to build off of the defined "id". Within the "$ref" I used the "/" prefix (you've defined this as a constant at RefValidator.REF_DOMAIN) in order to build off of the base domain path, but this is where the unexpected behavior comes into play.

If the "id" property is set to "http://localhost:9999", then I'd expect this "/location_schema.json" reference to resolve to "http://localhost:9999/location_schema.json". However, your library resolves this to "http://localhost/location_schema.json". This wouldn't be a problem if the port was 8080, but I'm using a nonstandard port in this situation. To _almost_ get the behavior that I originally expected, I'd have to change my reference to "/:9999/location_schema.json". However that also doesn't work because your library resolves that reference to "http://localhost//:9999/location_schema.json".

This is the json schema that I'm referring to:
```
{
  "type" : "object",
  "id": "http://localhost:9999",
  "properties" : {
    "age" : {"$ref" : "#/definitions/Age"},
    "location" : {"$ref" : "/location_schema.json"}
  },
  "definitions" : {
    "Age" : {
      "type" : "number",
      "minimum" : 0
    }
  }
}
```

Links:
https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/RefValidator.java
https://spacetelescope.github.io/understanding-json-schema/structuring.html#the-id-property

Wafflemon:
If you agree that this is an issue, then I would gladly submit a pull request containing a fix and a test case.

stevehu:
@Wafflemon Thanks for bringing it up. Yes. This is a problem as we are assuming the port number as default. It would be great if you could submit a pull request.

Wafflemon:
Hey @stevehu, I've been looking more into this issue and trying to fix some other issues that I could find in the RefValidator class as well. I ended up finding an existing test case that appears to be quietly failing. The problem lies within the src/test/resources/refRemote.json resource file. The "change resolution scope" schema at the end of the file is throwing an exception and causing this library to bypass validation for the associated tests.

Here's the exception that I'm getting:
```
[...]

Issue #55, opened 2017-12-19
Support for $ref external schema references?


harold-owen:
We have a schema that refers to three other schemas using $ref.  The others are at the same relative path (folder).  Trying to resolve by URL.  The first schema is:
https://scap.nist.gov/schema/nvd/feed/0.1/nvd_cve_feed_json_0.1_beta.schema

The others are
https://scap.nist.gov/schema/nvd/feed/0.1/cvss-v3.0.json
https://scap.nist.gov/schema/nvd/feed/0.1/cvss-v2.0.json
https://scap.nist.gov/schema/nvd/feed/0.1/CVE_JSON_4.0_min.schema

Getting warning in log:
16:27:08,402 WARN  [com.networknt.schema.JsonMetaSchema] (quartzScheduler_Worker-1) Could not load validator $ref

Any tips for troubleshooting?  Is $ref supported by json-schema-validator?  Nice library BTW, it has saved us a lot of time.

stevehu:
{"$ref": "CVE_JSON_4.0_min.schema"} cannot be resolved normally as json schema reference should be a uri. If you are using the uri (https://scap.nist.gov/schema/nvd/feed/0.1/CVE_JSON_4.0_min.schema) it should work. If you want to refer to local filesystem, you need to take a look at https://github.com/networknt/json-schema-validator/pull/45

harold-owen:
Thank you.  I will try this and report back.

2beers:
Hello. I've used version 0.1.3 with $ref external schemas without any problem. Today I decided to upgrade to 0.1.13 and I received the following error:

```
Caused by: java.lang.NullPointerException
	at com.networknt.schema.RefValidator.obtainAbsolutePath(RefValidator.java:92)
	at com.networknt.schema.RefValidator.getRefSchema(RefValidator.java:60)
	at com.networknt.schema.RefValidator.<init>(RefValidator.java:45)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:488)
[...]

Issue #26, opened 2017-04-04
Schema id does not  declare a base URL for relative $ref URLs


xaviarias:
When resolving relative `$ref`erences, URLs are constructed using the reference value directly without using the base URL of the referencing schema.

https://spacetelescope.github.io/understanding-json-schema/structuring.html?highlight=dependencies#the-id-property

For example, if you own the `foo.bar` domain, and you had a schema for addresses, you may set its id as follows:

`"id": "http://foo.bar/schemas/address.json"`

This provides a unique identifier for the schema, as well as, in most cases, indicating where it may be downloaded. For example: `{ "$ref": "person.json" }`

A JSON schema validation library would fetch `person.json` from `http://foo.bar/schemas/person.json`, even if `address.json` was loaded from the local filesystem.

eskabetxe:
i was this issue too..
@stevehu i will check this..

eskabetxe:
created PR 
https://github.com/networknt/json-schema-validator/pull/37

it could be better, if we can had some level of cache for the url/content, it would not be necessary to be downloading the content to each $ref as it always made "new RefValidator"

stevehu:
@eskabetxe I think it is cached as we only parse schema in the factory once. And all the validators are created during parsing and cached in a map. I didn't look at the code but make the comment based on the way it was implemented. If you could confirm that we parse the ref on a per request basis, then we definitely need to get it fixed. Could you please confirm? Thanks.

eskabetxe:
for this example
`{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "id": "http://localhost:8080/testing.json",
  "properties": {
    "prop1": { "$ref": "rules.json#/prop1" },
    "prop2": { "$ref": "rules.json#/prop2" },
    "prop3": { "$ref": "rules.json#/prop3" }
  }
}`

the $ref file (rules.json) will be charge 3 times..
because for each prop we create a new RefValidator that will download rules.json and check the proper rule

i think that prop1, prop2, etc are cached, but the file rules.json is download 3 times and this could be improved

stevehu:
This issue should be resolved by #141

PR #141, opened 2019-05-03
Improved Ref Validator

jawaff:
This pull request's aim is to eliminate some of the limitations with the RefValidator class. 

Here are list of some of the limitations:
- Relative $ref required an absolute id to exist.
- The RefValidator#isRelativePath() check only took into account the 'http' protocol. (Relevant to #98 #26 #37).
- A $ref was only able to build off of a single id.
- A relative domain $ref only worked for absolute urls that followed a <protocol>:<host> pattern. (Relevant to #86).
- An id previously was required to have at least one "/" character. (Relevant to #55).

In order to solve these issues, I made it so that JsonSchemas are able to optionally store a 'currentUrl'. This currentUrl is effected by 'id' keywords, urls that are provided to the JsonSchemaFactory. The JsonSchemaFactory is operated by the user of this library as well as RefValidators that point to a remote schema. 

All of this then allows the RefValidator to easily deal with a relative $ref. We deal with the relative $ref by simply using the constructor of the URL class. The URL constructor is able to automatically handle domain and subdirectory relative references. The URL constructor just requires a parent URL (which is provided by the JsonSchema's currentUrl) and some relative/absolute string path. If the parent URL is null, then the constructor requires the string path to be an absolute url. If the parent URL is nonnull, then the string path can be either an absolute url or a relative path (absolute urls in the string path are able to override the parent URL). 

Looking at my changes though, the URL constructor isn't actually being used directly. I chose to use the existing URLFactory class in this project and added to it so that it can combine a parent URL with some string path. Doing this allows the RefValidator to support the 'classpath:' and 'resource:' protocols that were already working with this library.

Another thing to note is that I updated the interface for the JsonSchemaFactory. I essentially just duplicated the existing factory methods so that users can optionally provide the URL of the schema that they're dealing with. The JsonSchemaFactory is still backwards compatible and doesn't require the URL of the schema, but it can be supplied so that '$ref' and 'id' values are able to be relative to the schema's URL.

So in short, I basically updated the library so that it can rely on the URL constructor for the complicated logic of combining 'id' and '$ref' values with the url of a schema. The URL constructor basically handles all of what this validator needs to do (we even are able to add support for custom protocols) and that bit of code is builtin to Java, so I think this creates a robust and flexible solution.

Note1: I saw that unit tests (in the JsonSchemaTest file) with invalid schemas were succeeding, so I ended up making it so that those tests fail. There was one existing unit test that was silently failing and it has probably been silently failing in this project for a while. It had a reference to the src/test/resources/tests/folder/folderInteger.json file (which didn't exist). I figure that if a test schema isn't able to be loaded, then the unit test or the project probably needs to be fixed.

Note2: I fixed a unit test relating to issue #12 (i.e. SelfRefTest.java). It was loading a schema as a JsonNode (which was succeeding), but the test seemed like it should have loaded it as a JsonSchema. Since I fixed this unit test and it now correctly reproduces the problem in issue #12, the unit test now fails (as expected with a StackOverflow). I just added an @Ignore annotation until someone provides a fix.


jawaff:
It appears that Oracle's 1.8 JDK doesn't work with my fix. The URL constructor in that implementation seems to convert strings to lowercase when they are combined with a parent URL. This was not the case when I was testing with OpenJDK 11.

jawaff:
My development environment has a proxy, so some of the unit tests cause connection timeouts. I had some problems with a slight bug that I introduced into a schema, but I eventually figured out what the problem was. I initially thought that it had something to do with my Java version (I'm using OpenJDK 11 and Travis is using Oracle JDK 1.8), but it was just a simple typo in the end.

jawaff:
On a positive note, if I didn't add in that exception for the invalid schemas, then my previous commits would have passed the unit tests. Silently failing tests can be a bad thing potentially.

jawaff:
Anyway, I'm ready for this pull request to be reviewed. I'm done adding in the improvements that I was interested in. There's still some features that this library doesn't support regarding $refs, but I think it supports the main use cases and that seems fine to me.

jawaff:
I updated the JsonSchemaFactory a bit. I previously thought that I should add method overloads for everything so that users can associate a URL with some given schema String/InputStream. However, the user can just pass in a URL and then implement a custom URLFetcher to do whatever they need.

I did decide to keep the overload that our JsonSchemaTest currently is using. This particular use case isn't easily fixed by a URLFetcher because the schema can potentially exist within a larger json file, but may still need a URL to be associated with.

jawaff:
This project is once again up to date and ready to merge. I ran into an issue with the URL mapping while I was fixing a merge conflict in the RefValidator, but that is all worked out now.

stevehu:
@jawaff I have reviewed the PR and want another developer to review it before merging. The idea is to skip this weekend's release as well due to some emergency bug fixes in this library.

stevehu:
@jawaff We have got the emergency fix 1.0.10 out and going back to this PR. Could you please resolve the conflicts. Thanks a lot for your help.

stevehu:
@jawaff I have merged the PR and released version 1.0.14. Thanks a lot for your help. It is a huge PR and it takes some time to review. 

https://github.com/networknt/json-schema-validator/releases/tag/1.0.14

jawaff:
Hey Steve,

I thought I needed to make some updates. I'm on vacation right now, but I
can make any updates that are needed in a couple of weeks. Thanks for
checking out my updates.

stevehu:
No worry. I made a little adjustment after the merging and it works perfectly. Again, thanks for the help as this PR resolves a lot of pending issues.

flozano:
One Q, why using URL everywhere instead of URI? URL doesn't allow using URN values

jawaff:
Hello, a URI may define something like a book's ISBN and that makes no
sense in the context of defining the location of a file. A URL identifies
the resource and provides the information for accessing it.

I mostly went off of what the library was doing initially anyway. They
already had a classpath URL handler defined, so it was just convenient to
keep using URLs. URLs just make sense for this json schema validators
because they follow the same rules as the $ref.

As far as I can tell URLs meet all of the requirements and prevent things
that shouldn't be allowed (such as book ISBNs). URIs don't seem necessary
and aadditional URL schemes can be supported if need be.

flozano:
Why do you want to dictate that schemas should be resolv'able thru URLs? Some people can have schemas ( I do :) ) that may have different URIs resolvable thru different means. An URN is a more generic form that allows me to plug whatever I want to retrieve those. With some pluggability, it'd be very convenient to say that URIs in the form of "http://..." are resolved with your URL resolver, whereas URIs in the form of "urn:mycompany:...." are resolved in the way I "plug".

flozano:
(discussion ongoing in #166)

Issue #147, opened 2019-05-21
Fails to validate MIN and MAX when number type is converted to BigInteger


BalloonWen:
When parsing Json content, and convert numbers to BigInteger will cause MIN and MAX validation fail.

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [Fails to validate MIN and MAX when number type is converted to BigInteger](https://trello.com/c/OlhgMOth/223-fails-to-validate-min-and-max-when-number-type-is-converted-to-biginteger)

stevehu:
1.0.9 is released to resolve this issue. Thanks.

PR #148, opened 2019-05-21
Fix min/max validator

BalloonWen:
related issue: https://github.com/networknt/json-schema-validator/issues/147


stevehu:
Is this bug impacting any existing clients? Do we need a quick light-4j release to get it fixed?

BalloonWen:
It impacts one client, but I think for now just cut a release just for json-schema-validator is enough. They are dependent on this directly.

stevehu:
Thanks guys. I will release this library only tonight.

Issue #138, opened 2019-04-24
validation of date fields


pallaviterala:
Hi,

The date fields are not being validated correctly by framework. If an invalid date is provided in request, it's still not throwing any validation error.

e.g.: 2019-02-29

And for valid dates, the body map returned by BodyHandler is holding a String object instead of a Date object. If this map contains Date object, we can avoid parsing String to Date in APIs.

Regards,
Pallavi

stevehu:
In the JSON schema specification, the date is only validated with a regular expression to ensure that the format is correct. Given that we have enhanced this library a lot based on the requirement from our customers, I think it makes sense to implement additional validation for Date type. However, the extra validation might slow down the validation process for most users who have a valid date format coming from a standalone application or a Javascript SPA with a DatePicker. I would recommend making this additional validation configurable so that it can be enabled by the user if they have clients application that allows users to enter date as a string.

ddobrin:
Agreed with making it config-based

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [validation of date fields](https://trello.com/c/3VJNBPnt/238-validation-of-date-fields)

PR #149, opened 2019-05-21
Added DateTimeValidator

jiachen1120:
Related issue: https://github.com/networknt/json-schema-validator/issues/138

### DIFF
--- a/src/main/java/com/networknt/schema/DateTimeValidator.java
+++ b/src/main/java/com/networknt/schema/DateTimeValidator.java
@@ -56,7 +56,7 @@ public class DateTimeValidator extends BaseJsonValidator implements JsonValidato
             return errors;
         }
         if (!isLegalDateTime(node.textValue())) {
-            errors.add(buildValidationMessage(at, node.textValue()));
+            errors.add(buildValidationMessage(at, node.textValue(), formatName));
         }
         return Collections.unmodifiableSet(errors);
     }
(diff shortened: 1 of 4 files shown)


NicholasAzar:
@stevehu we should include this in the fix release too, just looking into a change quickly

stevehu:
It is OK. Let's get it reviewed and merged. We can have another release after this.

stevehu:
@jiachen1120 For most application, the date or date-time are generated from datepicker or generated from the client applications. These days, it is very rare that a string format of date can be passed to the API from the UI. I am wondering if you could make this additional date validation optional in the config file. It should be disabled by default and only be enabled if the user knows the API might be hit by invalid date.

NicholasAzar:
@stevehu even if the date is generated from a date picker or whatever client application.. I don't think you would ever want to blindly trust the client and allow invalid dates to be passed
Do you expect the performance overhead to be that significant?

stevehu:
@NicholasAzar The performance hit might not be significant but the date validation is not part of the JSON schema specification. We are enhancing this library to serve our framework; however, we should give users an option to use the library as a standard JSON schema validator. In our framework, we are using the configuration file to control the behavior already and we can set it enabled to true as default from light-codegen. For other users who are using the json-schema-validator as a standalone library, they would expect the standard validation defined in the spec.

NicholasAzar:
@stevehu I'm not sure I follow when you say that date validation is not part of json schema validation.. obviously we have the date & date-time formats, so you're saying that any numbers that fall into that format are automatically valid dates? So a valid date per json schema validation could be 2019-02-31 even though it's not a date in the real world?

NicholasAzar:
Also, from a quick search it does seem like date validation is in fact part of json schema from what I can tell.. json schema says dates should follow RFC3339 here: https://json-schema.org/draft-04/json-schema-validation.html#rfc.section.7.3.1.2

And RFC3339 on page 8 clearly indicates maximum day numbers for specific days here: https://www.ietf.org/rfc/rfc3339.txt

Am i missing something?

stevehu:
@NicholasAzar You are right. I was misled by the other implementations years ago when starting the library. Since there is an RFC for the date-time format, we should follow the RFC.

Issue #151, opened 2019-05-27
add validation for string type: uuid


chenyan71:
schema:
  type:string
  minLength: 1
  format: uuid

we need add new format validation for uuid format string

stevehu:
I found the following links and hope they are helpful. 

https://github.com/json-schema-org/json-schema-spec/issues/542

https://tools.ietf.org/html/rfc4122

https://stackoverflow.com/questions/50204588/how-to-define-uuid-property-in-json-schema-and-open-api-oas

chenyan71:
user can use pattern to define the uuid format. But since uuid is validate format for openAPI. We should have validation for it. Added separate validator for it

PR #152, opened 2019-05-28
fixes issue #151

chenyan71:
add new format validator for uuid:

format: uuid


stevehu:
@chenyan71 Could you please add test cases for this new validator? Thanks.

PR #159, opened 2019-06-05
Make a colon in date-time format zone optional (close #158)

### LINKED ISSUE #158: date-time format should consider colon in timezone optional
`2018-07-27T00:00:01.000-0300` date time is valid against all ISO8601/RFC3339 validators I tried, but this validator fails because colon is [necessary per regex](https://github.com/networknt/json-schema-validator/blob/2449ee0947c84f14872436f9ff85c806abbb0051/src/main/java/com/networknt/schema/DateTimeValidator.java#L41).

chuwy:


stevehu:
@chuwy Thanks a lot for your help.

chuwy:
Thank you, @stevehu and co for maintaining this library. We at snowplow switched to it recently from FGE JSON Schema validator and experience is so much smoother. We're validating tens of billions of JSON instances with very diverse schemas, so you can expect some corner case tickets from us soon :)

stevehu:
@chuwy Thanks for the good words. It is part of the light-4j service mesh and a lot of contributors are actively working on it. Let's work together for all the edge cases and make this library useful for more users.

Issue #160, opened 2019-06-05
when schema type is integer but max/min value is a float point number


BalloonWen:
whenever there's a combination like:
{"type": "integer",
"maximum": 1e+32}
will validate error

stevehu:
![](https://github.trello.services/images/mini-trello-icon.png) [when schema type is integer but max/min value is a float point number](https://trello.com/c/X3QwTNzO/255-when-schema-type-is-integer-but-max-min-value-is-a-float-point-number)

PR #164, opened 2019-06-11
Fix/#160 max value float

BalloonWen:
related issue: https://github.com/networknt/json-schema-validator/issues/160
did a performance test, it seems no big performance impact use BigDecimal to do the comparison, to use BigDecimal comparison will reduce a lot of defects.

tested with comparing with "1E39", "-1000" 200000 times after change. The average time is
2312ms
tested with comparing with "1000.1", "1000" 200000 times before change. The average time is 2326.8ms


src/test/java/com/networknt/schema/MaximumValidatorPerfTest.java
```diff
@@ -0,0 +1,62 @@
...
+public class MaximumValidatorPerfTest {
+    MaximumValidatorTest test = new MaximumValidatorTest();
+
+    @Test
+    public void testTime() throws InvocationTargetException, IllegalAccessException {
+        test.setUp();
+        String[] testMethodsToBeExecuted = {"testMaximumDoubleValue"};
+        List<Method> testMethods = getTestMethods(testMethodsToBeExecuted);
+        long start = System.currentTimeMillis();
+        executeTests(testMethods, 200000);
```
NicholasAzar: I don't think we should have performance tests be part of the build. Specially when there isn't an assertion. Could this be ignored?

BalloonWen: you're right. made it ignored.

Issue #166, opened 2019-06-18
Allow using URN and not just URLs


flozano:
(From https://github.com/networknt/json-schema-validator/pull/141)

Right now, if we understand correctly, references to external schemas are forced to be URLs.

We'd like to extend the validator to register some well-known URNs for some internal schemas, without publishing them in a valid URL, but we don't think it's possible in a clean way with current code-base, so if we want to use this library we need to replace those URNs to URLs in a pre-processing which shouldn't be required.

flozano:
related to https://github.com/networknt/json-schema-validator/issues/100

stevehu:
@flozano I agree that we need to support both. For now, most users are using external references as they want to share some comment schemas between the specifications of services. Eventually, users would like to refer to some common schemas.

flozano:
In our case, it's not so much about sharing with 3rd parties, but about our own private schemas. We have many schemas which use URNs that refer to a few well-known schemas. 

If we could use URI instead of URL it would be fixed. I could attempt to do a pull-request, but I'm afraid URL is scattered all over the code, so it'd be very big.

stevehu:
@flozano Please give it a try. We have sufficient test cases to ensure nothing is going to break.

stevehu:
@flozano Could you please provide several examples of URNs that are publicly accessible? Also, any documents or links that might help us to understand the requirement in your case.

flozano:
You can see one here: 
https://api.flightstats.com/flex/airports/rest/v1/schema/json

in this example, the ref refers to a URN inside the same document, but it could be externally defined. Basically I think schema validator should not dictate what references are and how they are fetched for everyone.

By using URL you enforce the references to be URLs, by using URI you can let the "fetcher" implementator decide how to resolve those.

flozano:
Please note that, as per https://json-schema.org/understanding-json-schema/structuring.html, $ref and $id are URIs and not URLs

stevehu:
For people (including me) who are confused about URL, URI, and URN, please check this link 

https://stackoverflow.com/questions/176264/what-is-the-difference-between-a-uri-a-url-and-a-urn

I think we should support all of them as they do have a place in the JSON schema.

jawaff:
[...]

PR #167, opened 2019-06-24
Fix/#166 uri support

jawaff:
To summarize what I did:
- Lots of term changes because we no longer deal with just URLs.
- The JsonSchemaFactory.Builder is now able to be configured with URIFactories and URIFetchers. These factories and fetchers are able to be associated with a number of URI schemes. 
- A default URIFactory and URIFetcher is configured for our custom Classpath URLs and standard URLs. (These can be overridden if the user wishes).
- The JsonSchemaFactory accepts a URISchemeFactory and URISchemeFetcher that are configured with the scheme->factory/fetcher maps. The factory and fetchers are chosen based on the URI scheme that is needed.
- The URL mapping feature should now work with URIs in general.
- URIs are created using the JsonSchemaFactory's URIFactory (which is the URISchemeFactory).
- A URI's stream is fetched with the JsonSchemaFactory's URIFetcher (which is the URISchemeFetcher).

I'll make any changes that are necessary.


jawaff:
I think I've got everything working. Let me know what you all think. We can talk design and maybe figure out a better solution if you'd like. This is mostly just a proposition and I'm open to changing the approach.

jawaff:
This pull request is meant for fixing issue #166. (I'm trying to get Github to link this to the issue).

PR #169, opened 2019-06-25
fix/#166 uri support update

jawaff:
@stevehu I just found a minor problem in the code that I had a pull request for. There was still a place where URI#create (from my earlier experiments) was being used instead of the URISchemeFactory#create. I also moved the URISchemeFactory into the ValidationContext, because it didn't seem to belong in the JsonSchemaFactory.


jawaff:
This should finalize the #166 issue.

PR #172, opened 2019-07-18
fixes #171 validation of quoted numerics added

### LINKED ISSUE #171: Support minimum/maximum on quoted numerals
## Context

As mentioned in #118 there is a "loose" mode support for numeric types. Although not part of the [JSON spec](https://www.json.org/), the likely reason for such mode is JavaScript's specification of [number type](http://www.ecma-international.org/ecma-262/9.0/index.html#sec-ecmascript-language-types-number-type) which basically uses only [64bit double type](https://en.wikipedia.org/wiki/Floating-point_arithmetic), aka 64-bit format IEEE 754-2008. Motivated by issues resulting from [this](https://www.techempower.com/blog/2016/07/05/mangling-json-numbers/) [well known](https://2ality.com/2012/07/large-integers.html) [behavior](https://blog.angularindepth.com/javascripts-number-type-8d59199db1b6) some languages converged on using serialized/quoted form when serializing long, double and bigger values to JSON, for example [protocol buffers](https://developers.google.com/protocol-buffers/docs/proto3#json)

Since quoted numeric values are accepted with loose typing other JSON Schema validators should handle them as well. 

This issue focuses on `maximum` and `minimum` validators at the moment.

## Sample test

A json payload like this
```json
"9223372036854775807"
```

should check out agains schema like these

**case 1**
```json
{
    "$schema":"http://json-schema.org/draft-04/schema#",
    "type": "integer",
    "maximum": 9223372036854775807
}
```

**case 2**
```json
{
    "$schema":"http://json-schema.org/draft-04/schema#",
    "type": "integer",
    "maximum": 1E+20
}
```

**case 3**
```json
{
    "$schema":"http://json-schema.org/draft-04/schema#",
    "type": "number",
    "maximum": "9223372036854775807.0"
}
```
**case 4**
```json
{
    "$schema":"http://json-schema.org/draft-04/schema#",
    "type": "number",
    "maximum": 922337203685477580700
}
```

Essentially either of `integer`/`number` types and any valid JSON number as `maximum`/`minimum` value (including quoted numeric type).

## Edge case

Not clear if the following combination should trigger a `maximum`/`minimum` validators
json
```json
"42"
```

schema
```
{
    "$schema":"http://json-schema.org/draft-04/schema#",
    "type": "string",
    "maximum": 100
}
```

kosty:
Proposed solution to issue raised in #171

## Testing

Organized tests into sections (positive and negative) tests for `number`, `integer`, and `integer` with exclusive type of behavior.

Each collection of inputs is augmented with it's corresponding quoted counter part. 

Quoting thresholds does not seem necessary as they can be parsed into numbers at schema creation time.

### DIFF
--- a/src/main/java/com/networknt/schema/MaximumValidator.java
+++ b/src/main/java/com/networknt/schema/MaximumValidator.java
@@ -48,20 +48,26 @@ public class MaximumValidator extends BaseJsonValidator implements JsonValidator
 
         parseErrorCode(getValidatorType().getErrorCodeKey());
 
+        final String maximumText = schemaNode.asText();
         if (( schemaNode.isLong() || schemaNode.isInt() ) && (JsonType.INTEGER.toString().equals(getNodeFieldType()))) {
             // "integer", and within long range
             final long lm = schemaNode.asLong();
             typedMaximum = new ThresholdMixin() {
                 @Override
                 public boolean crossesThreshold(JsonNode node) {
-                    long val = node.asLong();
                     if (node.isBigInteger()) {
                         //node.isBigInteger is not trustable, the type BigInteger doesn't mean it is a big number.
-                        if (node.bigIntegerValue().compareTo(new BigInteger(String.valueOf(Long.MAX_VALUE))) > 0) {
-                            return true;
-                        }
+                        int compare = node.bigIntegerValue().compareTo(new BigInteger(schemaNode.asText()));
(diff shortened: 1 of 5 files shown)


stevehu:
Version 1.0.17 has been released. Thanks a lot for the excellent collaboration. 

https://github.com/networknt/json-schema-validator/releases

Issue #173, opened 2019-07-19
AnyOfValidator ignores all previous validations errors if any of the type does not match


grssam:
Assume a schema with `anyOf` like the following:
```javascript
{
  "type": "object",
  "properties": {
    "a": {
      "anyOf": [
        {
          "type": "object",
          "properties": {
            "test": {
              "type": "string",
              "enum": [ "A", "B" ]
            }
          }
        },
        {
          "type": "null"
        }
      ]
    }
  }
}
```

Consider the following json being validated:

```
{"a": {"test": "C"}} 
```

When matched against both the sub-schemas in `anyOf`, it should give two error messages: one for enum failure for field `test` (schema `/properties/a/properties/test/enum`) and one for type mismatch for `null`

But the code in `AnyOfValidator.java#66`:

```java
        if (!expectedTypeList.isEmpty()) {
            return Collections.singleton(buildValidationMessage(at, expectedTypeList.toArray(new String[expectedTypeList.size()])));
        }
        return Collections.unmodifiableSet(allErrors);
```

basically makes the method ignore any other validation errors `allErrors` and only return the type mismatch error.

stevehu:
@grssam This is the traditional fail-fast vs fail-slow issue and since this library is used in the light-4j framework which encourages fail-fast as it is the trend for the modern cloud-native microservices. However, we understand different users would have different use cases and we want to cover all use cases. I think the best option would be to make this into the configuration to let users to choose fail-fast or fail-slow. What do you think?

grssam:
While I do agree with your ideology of fail-fast vs fail-slow, this issue does not below to that. When the code flow reaches the `if` block in the code present above, all the possible errors are already present in the `allErrors` variable because the code has already iterated over all the possible `schemas` surrounded by the `anyOf` operator.

stevehu:
@grssam Would you be able to submit a PR? Thanks.

grssam:
I would certainly help out here and fix the issue. I just wanted to get your opinion here.
So let me reconfirm here: The goal is that we return the errors present in `allErrors` along with the type mismatch error?

stevehu:
Yes. You are correct as this is just one particular validator. We should return all the errors for this particular validator. Thanks a lot for the clarification.

grssam:
Could you please assign it to me?

stevehu:
Done. I also invite you to the json-schema-validator team so that you can create a new branch for the issue.

stevehu:
[...]

PR #175, opened 2019-07-24
#173 - Do not ignore other errors in case of anyOf validator

grssam:
Fixes issue #173 
Also added a way to validate the error count in test cases using the optional int field `"errorCount"`


grssam:
(closed as the diff is coming out incorrect even after i fixed crlf vs lf)

stevehu:
@grssam It might be easier to create a branch instead of forking the repository to your organization.

grssam:
@stevehu Have sent another after fixing the CRLF issue. My IDE was auto converting to LF while committing.

grssam:
PS: I am not forking in my organization, but in my public user account. That's the ideal/suggested flow for OSS contribution (rather than using the actual repo's side branches) by outsiders :)

Issue #177, opened 2019-07-31
OneOf Validator Incorrectly Failing


jawaff:
I'm running into a problem with the OneOfValidator. I can't give a lot of information right now, but I'm going to be looking into this tonight.

The problem seems to be happening in this snippet:
```
if(numberOfValidSchema == 0){
    // one of has matched one of the elements
    // if it has an error here, it is due to an element validation error
    // within its child elements
    if(config.hasElementValidationError()) {
        errors.clear();
        errors.addAll(schemaErrors);
        break;
    } else {
        errors.addAll(schemaErrors);
    }
}
```

If we hit the `config.hasElementValidationError()` case, then the rest of the schemas in the oneOf are skipped. I have an example that is reaching this case, but the schema causing it shouldn't be triggering a match. It seems like we can just remove that case in order to fix my situation, but I'm not sure how that will impact other test cases.

I'm going to try reproducing in the unit tests and hopefully find out what is really happening. Then I'll get to making a pull request when I can.

Issue #111, opened 2019-01-03
Validation failure for optional field in a schema - in the PropertiesValidator 


ddobrin:
The JsonSchema class collects a number of validators which will be executed.
The RequiredValidator serves this purpose.

There are schemas with optional fields, used in oneOf, for example, where the field is not found and the properties validator considers it an error.

The RequiredValidator would exonerate that error, however it is called only later.

ddobrin:
![](https://github.trello.services/images/mini-trello-icon.png) [oneOf problem cascaded to validation - Validation failure for optional field in a schema - in the PropertiesValidator](https://trello.com/c/wFLqZCMV/60-oneof-problem-cascaded-to-validation-validation-failure-for-optional-field-in-a-schema-in-the-propertiesvalidator)

PR #178, opened 2019-08-01
Fix/#177 One Of Bug

jawaff:
This pull request is for issue #177.

First, I've added a commit that reproduces the problem. In essence, after encountering a schema failure, the remaining schemata in the OneOf array aren't being checked. I'm not entirely sure of the purpose of the code that I'm removing with this commit.

### DIFF
--- a/src/main/java/com/networknt/schema/OneOfValidator.java
+++ b/src/main/java/com/networknt/schema/OneOfValidator.java
@@ -152,16 +152,7 @@ public class OneOfValidator extends BaseJsonValidator implements JsonValidator {
                 errors = new LinkedHashSet<ValidationMessage>();
             }
             if(numberOfValidSchema == 0){
-            	// one of has matched one of the elements
-            	// if it has an error here, it is due to an element validation error
-            	// within its child elements
-            	if(config.hasElementValidationError()) {
-            		errors.clear();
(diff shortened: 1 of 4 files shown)


jawaff:
It looks like the fix was pretty simple. I just had to remove a little bit of code. I'm ready to merge this now.

BalloonWen:
this flag was initially introduced by @ddobrin , I do not quite understand the purpose of this flag either, but I think it's better to check with him if this is okay.
@ddobrin please advice.

ddobrin:
I think that you are referring to https://github.com/networknt/json-schema-validator/issues/111

My client has complicated specifications and the testcases in this modules did not catch the error at that time.
It was a showstopper and led to 1.0.1 if I am not mistaken.

The only way to validate that the respective functionality is not broken, and I do have APIs in production at this time, would be to build from @jawaff 's branch against the entire codebase and ascertain this. 

I will hold of on a review until I get a bit of time to do this.

stevehu:
@ddobrin Any update on your testing?

ddobrin:
Priority has 2 other PRs to look into beforehand :(

stevehu:
I have thoroughly reviewed the code and it makes perfect sense with the test cases. Merge it for the next release and will figure out if it breaks Dan's code.

PR #186, opened 2019-09-11
Issue185

ddobrin:


ddobrin:
Moved the complex validator from config to a separate ValidatorState ThreadLocal-based class

stevehu:
@yibogithub You are right. It is safe to use ThreadLocal.

PR #201, opened 2019-10-21
replace reflection in ValidatorTypeCode

### LINKED ISSUE #200: Use with obfuscation
Hi,
I use this library in Android and obfuscate the release build with Proguard. However, the library will crash after obfuscation due to the reflection code to instantiate validators inside the method `public JsonValidator newValidator(...)` in the class `ValidatorTypeCode`. All validator classes are renamed during compile time so they are not able to load using their original names during runtime. Could we replace the reflection code to make it compatible with obfuscation? Thank you.

complex1ty:
Fixes #200

### DIFF
--- a/src/main/java/com/networknt/schema/ValidatorTypeCode.java
+++ b/src/main/java/com/networknt/schema/ValidatorTypeCode.java
@@ -27,47 +27,47 @@ import com.fasterxml.jackson.databind.JsonNode;
 
 public enum ValidatorTypeCode implements Keyword, ErrorMessageType {
     ADDITIONAL_PROPERTIES("additionalProperties", "1001", new MessageFormat(
-            "{0}.{1}: is not defined in the schema and the schema does not allow additional properties")),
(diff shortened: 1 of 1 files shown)


stevehu:
@complex1ty All test cases are passed and the change makes sense and definitely better than the reflection. The only concern is that there are some items with null validator classes. I don't know if there is any possibility that some of the use cases from our clients will be broken. @ddobrin @chenyan71 @jiachen1120 what do you think? Thanks.

ddobrin:
@stevehu : I can run a test against the code at my client sometime later today and update the comments  here in case something is broken.

ddobrin:
I have used the branch for this PR and built the json-schema-validator locally.
Then rebuilt all APIs at my client.
I have not encountered any error.

While this was a necessary step in terms of validation, it might (or not) be sufficient.

stevehu:
1.0.22 has been released and the release note can be found at https://github.com/networknt/json-schema-validator/releases/tag/1.0.22

PR #202, opened 2019-10-24
Feat/#199 fail fast

khiftikhar:
Adding the feature to fail-fast once the very first validation error is discovered.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.networknt</groupId>
     <artifactId>json-schema-validator</artifactId>
-    <version>1.0.22</version>
+    <version>1.0.23</version>
     <packaging>bundle</packaging>
     <description>A json schema validator that supports draft v4</description>
     <url>https://github.com/networknt/json-schema-validator</url>
@@ -142,6 +142,7 @@
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
+                <version>4.2.1</version>
                 <extensions>true</extensions>
             </plugin>
             <plugin>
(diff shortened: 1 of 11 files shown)


stevehu:
@khiftikhar Sorry to review the PR late, I am working on the release of the light platform during this weekend. For the fail-fast flag, I see you have added to the ValidationContext which is an object that carries the runtime info from one validator to another. The best place for the fail-fast flag would be the SchemaValidatorsConfig class which is designed to add more configuration flags in the future. In this way, we can keep the signature of JsonSchema and JsonSchemaFactory as config object already passed in. This will make the API static without impact current users. What do you think? Thanks a lot for your help.

khiftikhar:
@stevehu I have moved the fail-fast flat to the SchemaValidatorsConfig.java now. Please take a look when you do have time.

stevehu:
@khiftikhar It looks very good to put the fail-fast flag into the config. This also gives us an example of how to add another flag in the future. I see that you are using an exception to throw the message when fail-fast is true. Is there any other way to return the error message when it happens in any validator? For some users, the exception might be treated as an alarm. I will look into it when I complete the release. Thanks.

khiftikhar:
@stevehu I had been looking into how we can stop the processing otherwise, but it did required a bit more changes into the code. 

I also thought to introduce a checked exception to `builder().failFast`method but then I thought it might make the usability a bit uglier.

Now, setting up this flag is a conscious decision of the user, and the documentation clearly specifies that no error message is returned, rather and exception is thrown. So it's the user's responsibility to take of it.

stevehu:
@khiftikhar You are right. This might be the easiest way to stop the validators. I think we need to start working on the documentation as the library is getting more and more features. Thanks a lot for your help.

stevehu:
@khiftikhar The 1.0.23 is released. Thanks a lot for your help.

PR #207, opened 2019-11-07
Fixing ref validation with same name but different files

andersonf:
Avoid java.lang.StackOverFlowError.
related issue https://github.com/networknt/json-schema-validator/issues/208

### DIFF
--- a/src/main/java/com/networknt/schema/RefValidator.java
+++ b/src/main/java/com/networknt/schema/RefValidator.java
@@ -45,6 +45,8 @@ public class RefValidator extends BaseJsonValidator implements JsonValidator {
     }
 
     static JsonSchemaRef getRefSchema(JsonSchema parentSchema, ValidationContext validationContext, String refValue) {
+        final String refValueOriginal = refValue;
+
         if (!refValue.startsWith(REF_CURRENT)) {
             // This will be the uri extracted from the refValue (this may be a relative or absolute uri).
             final String refUri;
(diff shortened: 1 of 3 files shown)


BalloonWen:
@andersonf Hi, I don't quite get the changes.  You change the "refValue" to "refValueOrigin", what issue it will fix? if you have created a related issue, can you attache the link in your description? also I don't get it how it avoid StackOverFlowError based on your description.
Thank you.

andersonf:
@BalloonWen Use the test I added in this branch but run it into the release/1.0.25 without change into refValidation. Then you will understand the issue.

andersonf:
@BalloonWen 
What I understood we have a Map to do a control of something (refParsingInProgress).
The key of map is a String(refValue):
#/definitions/c
#/definitions/b
#/definitions/a

The issue appers at line 73: refValue = refValue.substring(index);

As you can see in the test I`m using the same attribute into the object "a" but in different files.
At line 73 the code will the reformat the value of "a" attribute
Value: "subSchemas.json#/definitions/a"
to: "#/definitions/a"

But the value "#/definitions/a" already inserted into the Map, and it will do an eternal loop.
To fix this issue is just to save into the key the complete name of refValue.
"subSchemas.json#/definitions/a"

BalloonWen:
I see, so it lost the original value because of the substring.
Thank you for your detailed explanation.

BalloonWen:
Hi @andersonf, PRs should be tight to issues, I created an issue based on this, please see if it's accurate, if not please feel free to modify it. 
https://github.com/networknt/json-schema-validator/issues/208

stevehu:
@andersonf @BalloonWen @jiachen1120 I have just released 1.0.26 with this PR. Thanks a lot for your help.

PR #213, opened 2019-11-15
fixes #54 support for draft V6, V7 and V2019-09

### LINKED ISSUE #54: support for draft V6, V7 and V2019-09
I am looking to use constant (https://github.com/json-schema-org/json-schema-spec/issues/58) from draft v6 and there is a proposal to use Switch, so would there be any chance of supporting for draft v6 in future?

stevehu:
I have spent several hours today to refactor the library so that we can support multiple specification versions easily for both developers and end-users. @jawaff and I have discussed the design for several days in the https://github.com/networknt/json-schema-validator/pull/206 and we finally decide to implement it this way. 

I also added a document in the doc folder named specversion.md for both users and developers.


src/main/java/com/networknt/schema/JsonMetaSchema.java
```diff
@@ -38,41 +38,47 @@
...
-    private static class DraftV4 {
+
+    static PatternFormat pattern(String name, String regex) {
+        return new PatternFormat(name, regex);
+    }
+    public static final List<Format> BUILTIN_FORMATS = new ArrayList<Format>();
+
+    // this section contains formats that is common for all specification versions.
+    static {
+        BUILTIN_FORMATS.add(pattern("time", "^\\d{2}:\\d{2}:\\d{2}$"));
```
jawaff: It looks like we need to investigate the builtin formats a bit. The 'time' format isn't technically apart of draft 4 and there are other formats that we're not yet supporting such as 'duration'. So far the time-related formats are all backwards compatible since they follow RFC 3339 section 5.6. However, I'm worried that the formats aren't always backwards compatible, but I can't find any evidence of that yet. I wonder if it would be worth having different sets of builtin formats based on the chosen draft version.

draft 4 builtin formats:
https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-7.3 

draft 2019-09 builtin formats:
https://tools.ietf.org/html/draft-handrews-json-schema-validation-02#section-7.3

jawaff: I might have spoken too soon. Looks like you already started to allow draft specific formats. There is however a potential issue with the BUILTIN_FORMATS being shared by all of the V# classes. If the V4 and V6 classes both have their static blocks executed, then the BUILTIN_FORMATS will become polluted with both draft specific formats. That's a corner case that probably won't be encountered, but maybe someone would want to do it for whatever reason.

An easy remedy would be to have a static 'BUILTIN_FORMATS' variable in each of the V# classes and to utilize the formats from the previous draft version. This allows a single application to utilize JsonSchemas for different draft versions without having unexpected behavior. Our testing environment could have probably run into this problem since we're going to theoretically have to test all of the various draft versions.

Following is a quick snippet of what I was talking about. I assumed that builtin formats are backwards compatible, but I'm not entirely sure about that.
```
class JsonMetaSchema {
    public static final List<Format> BUILTIN_FORMATS = new ArrayList<>();
    static { ... }

    class V4 {
        public static final List<Format> BUILTIN_FORMATS = new ArrayList<>(JsonMetaSchema.BUILTIN_FORMATS);
        static { ... }
    }

    class V6 {
        public static final List<Format> BUILTIN_FORMATS = new ArrayList<>(JsonMetaSchema.V4.BUILTIN_FORMATS);
        static { ... }
    }
}
```

jawaff: Another thing, would it be a good idea to make an immutable list for the BUILTIN_FORMATS? I'm not sure if people do it, but it's technically possible for users to add to the BUILTIN_FORMATS. That might be a bad thing because it allows users to shoot themselves in the foot on accident. (Java 9's List.of() creates an immutable list if you want to use that.)

src/test/java/com/networknt/schema/BaseJsonSchemaValidatorTest.java
```diff
@@ -49,7 +49,7 @@ protected JsonNode getJsonNodeFromUrl(String url) throws Exception {
     }
 
     protected JsonSchema getJsonSchemaFromClasspath(String name) throws Exception {
-        JsonSchemaFactory factory = JsonSchemaFactory.getInstance();
+        JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
```
jawaff: I'm away from my IDE, so I don't know what the subclasses are of this class. Anyway, wouldn't it be better to allow the SpecVersion to be passed in through a protected constructor? That would allow the subclasses to easily choose the spec version they need to use. The other option is to make the SpecVersion an additional argument to these methods.

jawaff:
The changes look really good for the most part. The only other thing I can think to comment on is the fact that the draft 4 SpecVersion is hard coded into a lot of the tests. Maybe you could use Junit's Parameterized test runner. (I use TestNG more than Junit, so sorry if I'm not sure how to use this correctly.) It looks like the Parameterized test runner would allow the tests to be executed for each SpecVersion. That's just a suggestion in order to improve the tests, but it might not be necessary.

src/main/java/com/networknt/schema/JsonMetaSchema.java
stevehu: You are right. I thought there is only one instance of the JsonMetaSchema that needs to be created, but we cannot guarantee that users won't create another instance. I will make the change.

stevehu: I think we want to BUILDIN_FORMATS to be open so that customized formats can be injected easily. I haven't thought about how this can be done yet.

src/test/java/com/networknt/schema/BaseJsonSchemaValidatorTest.java
stevehu: The JsonShemaFactory is a singleton at the moment. It is designed this way to prevent multiple instances to be created. I am open to making it more flexible for users.

stevehu:
I haven't thought about the test cases in detail yet. Hardcoded V4 is to ensure that all the changes are backward compatible so that it won't impact the current users. If there is a way that we can run the test against all spec factory, it will be great.

src/test/java/com/networknt/schema/BaseJsonSchemaValidatorTest.java
jawaff: I get that it's a singleton. I was talking about subclasses of the BaseJsonSchemaValidatorTest. I was meaning something like the following, so that dependency injection is possible:

```
class BaseJsonSchemaValidatorTest {
  private final SpecVersion.VersionFlag version;
  protected BaseJsonSchemaValidatorTest(SpecVersion.VersionFlag version) {
    this.version = version
  }

  protected JsonSchema getJsonSchemaFromClasspath(String name) throws Exception {
    JsonSchemaFactory factory = JsonSchemaFactory.getInstance(this.version);
    ...
  }
}
```

src/main/java/com/networknt/schema/JsonMetaSchema.java
jawaff: The JsonMetaSchema.BUILTIN_FORMATS is public and static, so any user can very easily inject custom formats. All it takes is something like:

```
class JsonSchemaFactoryFactory {
  public static JsonSchemaFactory create() {
    JsonMetaSchema.BUILTIN_FORMATS.add(new PatternFormat("new-keyword", "\\d\\d\\d"));
    // or
    JsonMetaSchema.BUILTIN_FORMATS.V7.add(new PatternFormat("new-keyword", "\\d\\d\\d"));

    // The BUILTIN_FORMATS needs to be updated prior to the creation of the JsonSchemaFactory. JsonMetaSchema will utilize a completely new formats list after it is created with the JsonMetaSchema.Builder (which is great and I like it that way).
    return JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
  }
}
```

jawaff:
I don't know if I should commit to your pull request. I was always taught not to commit to other people's branches at my workplace. I'm just going to show an example of what I'm talking about and I can add a commit if you really want me to.

```
// This tells Junit that this test class should be parameterized.
@RunWith(Parameterized.class)
public class UriMappingTest {
    // This specifies the static method that supplies the parameters for each execution.
    @Parameterized.Parameters
    public static Collection<?> parameters() {
      return Arrays.asList(new Object[][] {
         { "draft4", JsonMetaSchema.getV4() },
         { "draft6", JsonMetaSchema.getV6() },
         { "draft7", JsonMetaSchema.getV7() },
         { "draft201909", JsonMetaSchema.getV201909() }
      });
    }

    private final ObjectMapper mapper = new ObjectMapper();
    private final ClasspathURLFactory classpathURLFactory = new ClasspathURLFactory();
    private final URLFactory urlFactory = new URLFactory();

    private final String resourcePrefix;
    private final JsonMetaSchema metaSchema;
    // This constructor is where the parameters are injected by Junit.
    public UriMappingTest(String resourcePrefix, JsonMetaSchema metaSchema) {
        super()
        this.resourcePrefix = resourcePrefix;
        this.metaSchema = metaSchema;
    }

   // This test will run 4 times. Each time will be with different this.resourcePrefix and this.metaSchema values.
    // One assumption is that there is a uri-mapping.json file in each one of the draft resource directories. It's possible that the same uri-mapping.json file could be reused for each draft version to ensure backwards compatibility, but maybe the uri-mapping.json files need to be different depending on the draft version.
   @Test
    public void testBuilderUriMappingUri() throws IOException {
        URL mappings = ClasspathURLFactory.convert(
                this.classpathURLFactory.create(String.format("resource:%s/uri_mapping/uri-mapping.json", this.resourcePrefix)));
        Builder builder = JsonSchemaFactory.builder()
                .defaultMetaSchemaURI(this.metaSchema.getUri())
                .addMetaSchema(this.metaSchema)
                .addUriMappings(getUriMappingsFromUrl(mappings));
        JsonSchemaFactory instance = builder.build();
        JsonSchema schema = instance.getSchema(this.urlFactory.create(
                String.format("https://raw.githubusercontent.com/networknt/json-schema-validator/master/src/test/resources/%s/uri_mapping/uri-mapping.schema.json", this.resourcePrefix)));
        assertEquals(0, schema.validate(mapper.readTree(mappings)).size());
    }
}
```

stevehu:
@jawaff I don't mind if you check into the same branch and this might be the most efficient way to work together. However, if you don't feel comfortable, could you please approve the PR and you can start another one after the merge. Thanks.

src/test/java/com/networknt/schema/BaseJsonSchemaValidatorTest.java
stevehu: Yes. It works much better and flexible.

stevehu:
@andersonf @jawaff Thanks a lot for the reviews and feedbacks. I have merged this PR. Please submit the PRs for the test runner and if-else-then. Contribution credits need to go to the contributors :)

jawaff:
@stevehu This was such a great change. Thanks for the effort, Steve. I'll look into making a pull request later today or tomorrow.

stevehu:
@jawaff I was inspired by the ideas you suggested.

PR #218, opened 2019-11-19
Added an example code snippet for yaml validation

piyushkumar13:
Added an example code snippet for yaml validation


stevehu:
I just realized that your use case is different than mine. We are using openapi.yaml for the schema and the data is coming from the request or response in JSON format. Either way, it works. Thanks a lot for the contribution.

PR #219, opened 2019-11-19
Fix for oneOf when not all properties are matched

aznan2:
The schema
```
"oneOf": [
  {
    "properties": {
      "bar": {"type": "integer"},
      "baz": {"type": "integer"}
    },
    "required": ["bar"]
  },
  {
    "properties": {
      "foo": {"type": "string"}
    },
    "required": ["foo"]
  }
]
```
fails to match the instance 
```
{"bar": 2}
```
because `PropertiesValidator` exits the first oneOf branch when "baz" is not matched, even though it is not required. This PR fixes this.

### DIFF
--- a/src/main/java/com/networknt/schema/PropertiesValidator.java
+++ b/src/main/java/com/networknt/schema/PropertiesValidator.java
@@ -48,7 +48,7 @@ public class PropertiesValidator extends BaseJsonValidator implements JsonValida
         	state = new ValidatorState();
         	validatorState.set(state);
         }
-        
+
         for (Map.Entry<String, JsonSchema> entry : schemas.entrySet()) {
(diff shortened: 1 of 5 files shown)


stevehu:
@aznan2 Your test case makes perfect sense in my opinion. As we constantly sync with the official test suites, the modification for the oneOf.json might be overwritten easily. Would you please open an issue in https://github.com/json-schema-org/JSON-Schema-Test-Suite?

aznan2:
@stevehu Sure! Filed as https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/307.

stevehu:
@aznan2 Thanks for opening an issue at json-schema-org to add the test case. I am monitoring the progress there and let's finalize our test cases once the dust is settled.

PR #220, opened 2019-11-21
Feat/multi-version-tests

jawaff:
This pull request is all about improving the existing unit tests. The current unit tests are mainly using the draft 4 version, but we should be making sure that these unit tests are ran for all of the compatible draft versions. By doing this we can guarantee backwards compatibility between draft versions.

### DIFF
--- a/src/test/java/com/networknt/schema/BaseJsonSchemaValidatorTest.java
+++ b/src/test/java/com/networknt/schema/BaseJsonSchemaValidatorTest.java
@@ -27,6 +27,12 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  * Created by steve on 22/10/16.
  */
 public class BaseJsonSchemaValidatorTest {
+    protected final SpecVersion.VersionFlag specVersion;
+    
+    protected BaseJsonSchemaValidatorTest(final SpecVersion.VersionFlag specVersion) {
+        this.specVersion = specVersion;
+    }
+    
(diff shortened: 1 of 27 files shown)


jawaff:
I'm not ready yet, but I thought I'd open up a draft pull request so you guys know that I didn't forget about this feature.

jawaff:
I'm almost finished with the tests, but my development environment is throwing some strange jacoco error. However, my IDE can run the tests without a problem. This makes it difficult to catch the bugs I'm introducing. I'll get through this.

jawaff:
I may have uncovered a bug that only exists when the later draft versions are used. There's a couple of tests in the MaximumValidatorTest and MinimumValidatorTest classes that only fail when draft versions after v4 are used. I need to investigate some more, but this is a really good find potentially.

jawaff:
I'm going to have to pick this up another time. It's so close to done.

eirnym:
@jawaff Thank you for a PR to improve the tests. 

Could you update your work? 
Is there a chance to split the work and add a few smaller PRs rather than a huge one?

stevehu:
@jawaff @eirnym I am open to any option that can speed up the progress. Thanks.

fdutton:
We are using [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) to ensure compliance with the various specifications and most of the requested functionality is now present. Since this request is so old and has so many conflicts, it seems best to just close it.

PR #223, opened 2019-12-03
fixes #222 add document for URL mapping

### LINKED ISSUE #222: add document for URL mapping

stevehu:
@rhwood With a recent issue about the meta schema URL switching from HTTP to HTTPS, I have received so many requests from users on how to use the URL mapping. I have drafted a document without a lot of details in a branch and wondering if you could review and add more details about the usage. Thanks a lot for the good work and it is a very important feature of the library now.


stevehu:
@rhwood Cool. You know this the best. We just need to guide users to leverage this great feature with ease.

PR #231, opened 2019-12-06
Update description in pom.xml to match readme.md

reftel:


stevehu:
@reftel Thanks a lot for your help.

PR #243, opened 2019-12-30
Improve accuracy of rounding with multipleOf

seamusv:
Ensure the accuracy of the multipleOf contract.


stevehu:
@seamusv Thanks a lot for your contribution. This makes perfect sense and I am a little surprised that the official test case doesn't cover this scenario. Since we constantly sync the official test cases to our codebase, I think the best option for us is to separate the new test cases to another JSON file. At the same time, let's open an issue to add these to the official test suite. What do you think? 

https://github.com/json-schema-org/JSON-Schema-Test-Suite

seamusv:
@stevehu I can certainly make that request to the other repo.  Where would you suggest this test case be temporarily located since "optional" could also be refreshed?

seamusv:
@stevehu I have moved the test data out of the official specifications and into a top-level JSON file.  The test is only in the draft7 suite.  I'll submit a PR to the official test suite repo.

seamusv:
I have submitted the request at https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/321

stevehu:
@seamusv Thanks a lot for the effort. Whenever we have a separate file, it won't be overwritten.

Issue #221, opened 2019-11-28
Android 6 support


matthias-sattel:
Hi,
in order to use your library on Android 6 (API>=23) we had to backport it to Java 6. Would this be a contribution that you would like to integrate into this project or will you stick to Java 8 ?
Best regards,
Matthias

stevehu:
@msattel I am not an Android developer and recently upgrade to Java 8 in order to leverage the new Java 8 features. I did a research and found Android 6 was discontinued in 2018 with on security patches anymore. Do we still have developers working on this platform?

stevehu:
https://github.com/networknt/json-schema-validator/issues/162

msattel:
@stevehu As far as I know around 15% of the Android users are still running on Android version 6 and below. This is why our management devided, that we have to support those old Android versions. Otherwise we would not support it at all and hopefully this situation will change soon.

stevehu:
@msattel I tried to change the java version back to 1.6 as below and found only a few places need to be updated in the src. There might be some test cases leverage Java 8 lambda expressions. Would you like to submit a PR? 

```
        <java.version>1.6</java.version>
        <java.testversion>1.8</java.testversion>
```

miklish:
Android 1.6 is from 2009. That's a full 10 years ago. What kind of revenue are you expecting from users that can't afford to update their hardware at least once every decade?

msattel:
@miklish It is Anrdoid 6, not Android 1.6
https://en.wikipedia.org/wiki/Android_version_history#Android_6.0_Marshmallow_(API_23)
It is from 2015 and yes, outdated from my point of view. But seems to still have a lot of end users and thus possible app customers.

msattel:
@stevehu I will provide the PR asap. Have a nice weekend

ondrej:
Hi @msattel , are you still working on a PR for this? We just started using the library on Android all the way back to API level 22 (Java 5.1) and while we haven't encountered any issues yet, would be keen to see what could break and if we can help make it official. Thanks!

msattel:
@stevehu @ondrej  Sorry for the delay, I just created the PR https://github.com/networknt/json-schema-validator/pull/244

stevehu:
@msattel Thanks a lot. I will mark this issue closed.

stevehu:
[...]

PR #244, opened 2020-01-03
Feature/221

msattel:
This pull request is about Android 6 support (API level >=23) as discussed in https://github.com/networknt/json-schema-validator/issues/221


stevehu:
@msattel I saw you have set up two different releases to the maven central. I am not familiar with this approach and wondering if you could point me to some documentation or guidelines. In my opinion, we can just downgrade the main release to Java 1.6 and it will work with all environments. What do you think?

msattel:
@stevehu  Sorry, do you mean the changes to the testsources java language level?
Downgrading the code level to 1.6 should work with the changes, so probably having separate settings for the test code is obsolete

stevehu:
@msattel In my opinion, we don't need to do anything with the test folder as they are not part of the release package on the maven central. What we can do is to ensure that src is compiled with JDK 1.6 and test is compiled with JDK 1.8. What do you think?

```
        <java.version>1.6</java.version>
        <java.testversion>1.8</java.testversion>

```

msattel:
@stevehu sound good, from my point of view

stevehu:
@msattel Would you be able to make the adjustment for the PR based on our discussion? Thanks.

stevehu:
@msattel Thanks a lot for your help. Do you know if the binary will be working if I am building with Java 8? I assume it will work but I need somebody to help to confirm.

msattel:
@stevehu thanks a lot for merging it

stevehu:
@msattel Thanks for the confirmation. The codebase is in Java 6 now even the test as some browser cannot separate between main and test. The library is released with JDK8 and I assume it should work.

PR #247, opened 2020-01-09
Resolve schema id from the schema document (for v6 and above)

martin-sladecek:
Schema id has not been resolved correctly for v6 and up as there's different id field name.

Not sure if you require to create a new issue for every pull request, seems like the problem is known (there were ignored test cases), but I haven't found any issue related to this.


stevehu:
@martin-sladecek Thanks a lot for the contribution. I realized the issue but haven't figured out the root cause yet due to other tasks. I've created an issue and linked to this PR as I found most people are looking at the issues instead of looking into the PRs.

PR #252, opened 2020-02-20
Issue-226 ContainsValidator

Asamsig:
I have implemented the ContainsValidator, I based it mostly on the ItemsValidator.
I noticed we already had tests for this, so I simply uncommented these.

Based on the tests, if the validation is attached to something not an array, we simply skip the validation, I guess this is fine?

### DIFF
--- a/src/main/java/com/networknt/schema/ContainsValidator.java
+++ b/src/main/java/com/networknt/schema/ContainsValidator.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2016 Network New Technologies Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.networknt.schema;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collections;
+import java.util.Set;
+
+public class ContainsValidator extends BaseJsonValidator implements JsonValidator {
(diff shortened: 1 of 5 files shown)


src/main/java/com/networknt/schema/ContainsValidator.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (node.isEmpty()) {
+            // Array was empty
+            return buildErrorMessageSet(at);
+        } else if (node.isArray()) {
+            boolean contains = false;
+            int i = 0;
+            for (JsonNode n : node) {
+                contains = contains || schema.validate(n, rootNode, at + "[" + i + "]").isEmpty();
+                i++;
```
Asamsig: Perhaps adding a `if(contains) break;` would be ideal, to decrease interations.

stevehu: For most of the scenarios, it doesn't matter as the number of elements won't be big. However, we are trying to provide a generic library and we cannot assume the use cases from users. There might be some crazy use cases with too many elements so I think you are right about to add the break.

Asamsig: I have pushed a new version, where I made it short circuit on the first match. This also somewhat simplifies the code.

src/main/java/com/networknt/schema/ValidatorTypeCode.java
```diff
@@ -71,7 +71,8 @@ public JsonValidator newValidator(String schemaPath, JsonNode schemaNode, JsonSc
     EXCLUSIVE_MINIMUM("exclusiveMinimum", "1039", new MessageFormat("{0}: must have a exclusive minimum value of {1}"), ExclusiveMinimumValidator.class, 14),
     TRUE("true", "1040", null, TrueValidator.class, 14),
     FALSE("false", "1041", new MessageFormat("Boolean schema false is not valid"), FalseValidator.class, 14),
-    CONST("const", "1042", new MessageFormat("{0}: must be a constant value {1}"), ConstValidator.class, 14);
+    CONST("const", "1042", new MessageFormat("{0}: must be a constant value {1}"), ConstValidator.class, 14),
+    CONTAINS("contains", "1043", new MessageFormat("{0}: did not contain an element that passes these validations: {1}"), ContainsValidator.class, 14);
```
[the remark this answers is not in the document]
Asamsig: Better to be consistent, I'll update the PR. :)

stevehu:
@Asamsig I have released 1.0.31 with the contains validator. Thanks a lot for your contribution. 

https://github.com/networknt/json-schema-validator/releases/tag/1.0.31

PR #259, opened 2020-03-01
validation fails with StackOverflowError

mihaiba:
validation fails with StackOverflowError when scheme with cyclic dependencies are used

### DIFF
--- a/src/main/java/com/networknt/schema/ValidationContext.java
+++ b/src/main/java/com/networknt/schema/ValidationContext.java
@@ -16,11 +16,12 @@
 
 package com.networknt.schema;
 
-import java.util.HashMap;
-import java.util.Map;
-
 import com.fasterxml.jackson.databind.JsonNode;
 import com.networknt.schema.uri.URIFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class ValidationContext {
     private final URIFactory uriFactory;
@@ -28,6 +29,7 @@ public class ValidationContext {
     private final JsonSchemaFactory jsonSchemaFactory;
     private SchemaValidatorsConfig config;
     private final Map<String, JsonSchemaRef> refParsingInProgress = new HashMap<String, JsonSchemaRef>();
+    private static final Map<String, JsonValidator> validatorCache = new ConcurrentHashMap<String, JsonValidator>();
 
     public ValidationContext(URIFactory uriFactory, JsonMetaSchema metaSchema, JsonSchemaFactory jsonSchemaFactory, SchemaValidatorsConfig config) {
         if (uriFactory == null) {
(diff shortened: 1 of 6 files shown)


stevehu:
@mihaiba Have you tried to debug into it and find the root cause? @jawaff What do you think about this test case? Thanks.

mihaiba:
Since there is a cyclic dependency between schemes I've created a cache for the validators that are being created based on the schema path. The test I've created it's passing with this changes, others are failing, any hints on this issue @stevehu @jawaff?

stevehu:
I saw the failed auto build. Let me take a look at the code from your forked repo. Thanks for the effort.

stevehu:
In the test cases, you have three JSON files in the cyclic folder of draft4. Is it possible to merge them into one file? JSON schema doesn't support local filesystem reference.

francesc79:
@mihaiba @stevehu  I have the same problem... I checking the PR but with a validatorCache inside a ValidationContext doesn't resolve the problem.
The cache return the same validator for all properties because the key is schemaPath. The schema path is the same for all properties. You are agree about that?

francesc79:
The cycle loop is created because the same Validators in a constructor create a new JsonSchema. In the JsonSchema constructor you  have 
```
this.validators = Collections.unmodifiableMap(this.read(schemaNode));
```
this code creates a loop.
In my opinion, the validators should have an initialization phase executed when all validators are costructed. The cache key should have a concatenation of currentURI, schemaPath and keyword. Do you agree?

stevehu:
@francesc79 I think you go the point. I didn't debug into it yet, but I am wondering why the JsonSchema constructor is called more than one time. The only entry point is the newJsonSchema from JsonShemaFactory and it should be a singleton. The reason we have the JsonSchemaFactory is to prevent multiple instances of JsonSchema. Thanks.

francesc79:
@stevehu JsonSchema is created more times because in the schema we have $ref that create a loop.
In the @mihaiba Test you can see

```
{
  "id": "Extension.json",
  "type": "object",
  "datatype": true,
  "category": "fhir",
  "properties": {
    "url": {
      "type": "string",
      "format": "url"
    },
    "valueElement": {
      "$ref": "Element.json"
    }
  },
  "additionalProperties": false,
  "required": [
    "url"
  ]
}
```

```
{
  "id": "Element.json",
  "type": "object",
  "datatype": true,
  "category": "fhir",
  "properties": {
    "id": {
      "type": "string",
      "pattern": "[A-Za-z0-9\\-\\.]{1,64}"
    },
    "extension": {
      "type": "array",
      "items": {
        "$ref": "Extension.json"
      }
    }
  },
  "additionalProperties": false
}
```
Extension schema refer Element schema that refer Extension.

stevehu:
Yes. You are right. This line creates a new Schema object. I think the cache key you mentioned should work. Thanks a lot for the debugging and investigation.

https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/RefValidator.java#L67

mihaiba:
@stevehu & @francesc79 My second push does try create a cache for the validators, but somehow I broke other scenarios, so I think some work still needs to be done there

stevehu:
This PR is just for demo. Closing it.

Issue #249, opened 2020-01-20
Collecting information while validating the schema.


prashanthjos:

prashanthjos:
Is it possible to collect some kind of information about the data while we are validating the given data against the schema ? For example if I want to collect certain Id’s mentioned in the data against  an array of domain values defined in the schema and return it back to the caller of the validate method. Essentially I need to maintain a context object about my validation, fill values into it during validation  and collect it at the end of validation.

prashanthjos:
My understanding is to pass a context object across to all the implementations of validators, formats and keywords. This context object scope is limited to a single validation that is, it is reset after every validation of the Json data against the json schema. Starting with, this context object could be an extension to a simple map. We can let the API user create an instance of this context object and pass it to the validate method. Context object can collect information across all the formats, validators and keywords. Please let me know your thoughts.

stevehu:
Without a detailed design, I cannot imagine how it works. As you have suggested, we can allow users to pass in a context object that indicates which validator will be used and what kind of data will be collected. In this case, we must check this context object in every validator to see if any data needs to be collected. This extra step would impact the performance across all validators in my opinion. Is there any better way that we can limit the impact? I think this is a useful extension for the library, but we need to carefully design it so that it won't impact most users. Would you be able to write something down or issue a PR so that we can have a base to discuss the details. Thanks.

prashanthjos:
Yes. I can fork out the repo and come up with a pull request. I am trying to make sure that code changes to the existing code is minimal and has no performance impact. I will get back to you with a PR. Thank you.

PR #260, opened 2020-03-06
Changes for adding collector context

prashanthjos:
These changes are with respect to issue #249 . Please let me know your comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@ For the latest version, please check the [release](https://github.com/networknt/
 
 ## [Customized MetaSchema](doc/cust-meta.md)
 
+## [Collector Context](doc/collector_context.md)
+
 
 ## Known issues
 
--- a/doc/collector_context.md
+++ b/doc/collector_context.md
@@ -1,0 +1,41 @@
+### CollectorContext
+
+
+There could be usecases where we want collect the information while we are validating the data. A simple example could be fetching some value from a database or from a microservice based on the data (which could be a text or a JSON object) in a given JSON node and the schema keyword we are using. 
+
+The fetched data can be stored some where so that it can be used later after the validation is done. Since the current validation logic already parses the data and schema, both validation and collecting the required information can be done in one go.
+
+CollectorContext and Collector classes are designed to satisfy this usage.
+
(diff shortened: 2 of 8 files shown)


stevehu:
@prashanthjos I looked at the code and had hard time understanding how it works. From the test case, it looks like you need to have customized validators to collect the data to the context. Are you expecting the users to provide the customized validators? Is it possible to write a document in the doc folder to let other users know how to leverage this new feature? Thanks.

prashanthjos:
@stevehu I will create a more sophisticated testcase and add a document. But quick explanation of my changes is I have created a singleton class CollectorContext.I have created a Collector interface which users can use to create/specify the objects they want to collect. These Collector objects are added to CollectorContext singleton while the JsonNode’s are being validated inside the validators. I have created a new method  validateAndCollect in JsonSchema class. The purpose of this method is to clear the CollectorContext object before validation and load the CollectorContext with Collector objects after validation. Since CollectorContext object is singleton designed using an enum it can be used anywhere inside the validators without changing any existing signatures. I have demonstrated that in the testcase. Let me know your thoughts.

prashanthjos:
There is no change to the signatures of existing Validators or other classes. Whenever someone needs to collect they get the CollectorContext singleton and add the Collectors to it.

stevehu:
Yes. I see there is no change to the signatures, which is very good. I just want other users to understand the implementation so that they can enjoy the new feature. Thanks a lot for the explanation and clarification.

prashanthjos:
I will come up with a doc and also what is the release cycle networknt follows.Do we merge the code into master branch and increment the version and a set of changes go together?

stevehu:
Yes. We can have another release as soon as it is merged.

prashanthjos:
Sure thank you

prashanthjos:
@stevehu I have changed the CollectorContext to be stored on ThreadLocal. I dint recognize the obvious thread contention issues with Singleton in my earlier commits. I have added some documentation and provided much clear test cases. Please review and let me know your thoughts.

stevehu:
@prashanthjos 1.0.32 is released moment ago. Thanks a lot for your contribution.

Issue #255, opened 2020-02-24
Dereferencing subschemas by $id with $ref in the same file does not seem to work


elzawie:
Hi there,

I am using your library for JSON validation and it works pretty well so far, however now I encountered a potential issue while trying to reference the internally defined subschemas using combination of $id and $ref properties. Referencing subschemas using JSON pointers instead of ids works as expected. The concept I am trying to test is described under the below hyperlink:

https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref

Using the same JSON schema and data file with the below online validators yield expected results - return error saying that the element of "cars" array is missing required properties. 

https://jsonschemalint.com/#!/version/draft-07/markup/json
https://www.jsonschemavalidator.net/

**JSON schema:**

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "$id": "https://graph.link/ecp/schema/CDF/garage.json",
    "type": "object",
    "properties": {
        "cars": {
            "type": "array",
            "items": {
                "$ref": "#car"
            }
        }
    },
    "definitions": {
        "car": {
            "$id": "#car",
            "type": "object",
            "properties": {
                "model": {
                    "type": "string"
                },
                "manufacturer": {
                    "$ref": "#manufacturer"
                }
            },
            "required": ["model", "manufacturer"]
        },
        "manufacturer": {
            "$id": "#manufacturer",
            "type": "object",
            "properties": {
                "name": {
                    "type": "string"
                },
                "factories": {
                    "type": "array",
                    "items": {
                        "$ref": "#factory"
                    }
                }
            }
        },
        "factory": {
            "$id": "#factory",
            "type": "object",
            "properties": {
                "location": {
                    "$ref": "#location"
                }
            }
        },
        "location": {
            "$id": "#location",
            "type": "object",
            "properties": {
                "country": {
                    "type": "string"
                },
                "city": {
                    "type": "string"
                },
                "employees": {
                    "type": "array",
                    "items": {
                        "$ref": "#employee"
                    }
                }
            },
            "additionalProperties": false
        },
        "employee": {
            "$id": "#employee",
            "type": "object",
            "properties": {
                "name": {
                    "type": "string"
                },
                "surname": {
                    "type": "string"
                }
            },
            "additionalProperties": true,
            "maxProperties": 3
        }
    }
}

```

**JSON data file:**

```json
{
    "cars": [
        {

        }
    ]
}

```

**Test code:**

```java

[...]

PR #263, opened 2020-03-09
255 issue id ref

rzukowski:
Fix for 'Dereferencing subschemas by $id with $ref in the same file does not seem to work' - issue https://github.com/networknt/json-schema-validator/issues/255

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
         </repository>
     </distributionManagement>
     <properties>
-        <java.version>1.6</java.version>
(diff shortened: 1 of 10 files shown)


stevehu:
@rzukowski Thanks a lot for your contribution. I have merged the changes into the master branch and created a test case for issue 255. The test case is failed due to some NPE issue. I am wondering if you could help to take a look as you are more familiar with the reference validator. The test case is in the issue255 branch. 

https://github.com/networknt/json-schema-validator/tree/issue255

PR #272, opened 2020-03-13
Use ECMA-262 validator when requested

### LINKED ISSUE #15: Regular expression should allowed input be limited?
Hello! I went back and read the regular expression section for the JSON schema standard. It turns out that a full ECMA 262 (aka Javascript) regular expression engine is not required. The specification says that implementations SHOULD follow ECMA 262 but not that it MUST use it. The specification also recommends that "schema authors SHOULD limit themselves to the following regular expression tokens..."

One possible feature to this library is to add validation to enforce that regular expressions are limited to the tokens that are recommended in the schema. This is one way to ensure compatibility between Java regular expressions and ECMA 262 regular expressions. The recommended tokens are:

* individual Unicode characters, as defined by the JSON specification [RFC7159];
* simple character classes ([abc]), range character classes ([a-z]);
* complemented character classes ([^abc], [^a-z]);
* simple quantifiers: "+" (one or more), "*" (zero or more), "?" (zero or one), and their lazy versions ("+?", "*?", "??");
* range quantifiers: "{x}" (exactly x occurrences), "{x,y}" (at least x, at most y, occurrences), {x,} (x occurrences or more), and their lazy versions;
* the beginning-of-input ("^") and end-of-input ("$") anchors;
* simple grouping ("(...)") and alternation ("|").

It would be a little tricky to implement the logic that only accepted the above tokens. Is it a good idea? What if someone really wants to use Java specific regular expressions? Should the library stop them?

eirnym:
Add ECMA-262 regex check support for `pattern`.

Fixes #15

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,6 +88,11 @@
             <version>${version.common-lang3}</version>
         </dependency>
         <dependency>
+            <groupId>org.jruby.joni</groupId>
+            <artifactId>joni</artifactId>
+            <version>2.1.31</version>
+        </dependency>
+        <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
             <version>${version.logback}</version>
--- a/src/main/java/com/networknt/schema/PatternValidator.java
+++ b/src/main/java/com/networknt/schema/PatternValidator.java
@@ -17,6 +17,11 @@
 package com.networknt.schema;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import org.jcodings.specific.UTF8Encoding;
+import org.joni.Option;
+import org.joni.Regex;
+import org.joni.Syntax;
+import org.joni.exception.SyntaxException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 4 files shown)


stevehu:
Cool. Thanks a lot for your contribution.

stevehu:
I have released 1.0.35 and added a document on how to choose the regex library. Please review the document and make changes if necessary. Thanks. 

https://github.com/networknt/json-schema-validator/blob/master/doc/ecma-262.md

SridharSubramaniam:
It would be good to add these new additions as separate repo instead of adding dependencies to the core lib. (especially when it requires 3rd party lib and is an optional feature)

Adding dependencies would need lot of security reviews in a large organization and it slows down the adoption of new releases. Although we won't use this new lib, we had to get it reviewed as it is now a dependency for this core validator.

eirnym:
@SridharSubramaniam per standard the library should use ECMA-262 compatible regex, while `java.util.regex` is quite buggy and not compliant to standard. 

That library provide compliance and everything is quite good. 

For example, `java.util.regex` successfully compiles and matches nothing this regex: `{2,3}` (yes, I've omitted a token to repeat), and this could be not the only issue.

BTW, it's theoretically possible to use `provided` library to fulfil your requirement before this behaviour become default.

@stevehu Please, make a decision, I'll provide a patch for it if needed.

stevehu:
@SridharSubramaniam I totally understand your concerns and want to make sure that can help to smooth the upgrade process. As @eirnym mentioned, it is OK with your organization if we change the dependency to `provided`?

eirnym:
@stevehu Yes, thank you

SridharSubramaniam:
@stevehu I submitted 1.0.36 for review - waiting for review. Will update based on the outcome.. probably a couple of days out. Will get back. thanks!

PR #289, opened 2020-04-28
Adding getAll method on CollectorContext class.

prashanthjos:
I have a use case where I need to get all the keys that were added to CollectorContext and operate on the values together for all these keys. Added a getAll method which gives me all the data collected by CollectorContext object. Please review.


stevehu:
@prashanthjos merged and released. Thanks.

https://github.com/networknt/json-schema-validator/releases/tag/1.0.39

PR #323, opened 2020-08-24
FIX: potential duplicate log entry due to race condition

kkonrad:
There was potentially race condition that could lead to situation in which we could get multiple log entries regarding the same keyword being unknown. Such case was when more than one thread was doing check on `containsKey` before they went into `put`.

Simple fix for this is to always add element to the `UNKNOWN_KEYWORDS` map and than check if it was present. See [documentation for `put` method and what value it returns](https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ConcurrentHashMap.html#put(K,%20V)). Doing this that way shouldn't impact performance.


stevehu:
@kkonrad Thanks a lot for your help.

PR #324, opened 2020-09-16
Generate module-info, fix build on JDK11+

handcraftedbits:
This PR makes `json-schema-validator` work better with JDK9+ w.r.t. modularization and a couple build fixes for JDK11+:

* `commons-lang` is bumped from `3.5` to `3.7` so the generated `module-info` can point to a fixed module name (instead of using an "automatic module name" determined from the JAR's filename, which is a discouraged practice).  `slf4j-api` was not bumped because the next version with any module support is `2.0.0-alpha1`, which I figured was too drastic of a version bump.
* `moditect-maven-plugin` is added to build a `module-info.class` class file which serves as the module descriptor.  This works even during a JDK8 build because `moditect-maven-plugin` will generate the `.class` file itself and place it under `META-INF/versions/9`, making this a [multi-release JAR](https://openjdk.java.net/jeps/238).  JDK8 will ignore `module-info` at runtime and JDK9+ will make use of it.
* `maven-javadoc-plugin` and `jacoco-maven-plugin` had their versions bumped to support JDK11+.

Edit: when I said JDK8, that's the lowest version I tested on.  I just noticed that JDK6 is the default build target for this project... I believe this should still work fine with the multi-release JAR trick.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,7 +63,7 @@
         <java.testversion>1.8</java.testversion>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <version.jackson>2.10.4</version.jackson>
-        <version.slf4j>1.7.25</version.slf4j>
+        <version.slf4j>1.7.30</version.slf4j>
         <version.common-lang3>3.7</version.common-lang3>
         <version.joni>2.1.31</version.joni>
         <version.logback>1.2.3</version.logback>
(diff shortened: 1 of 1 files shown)


stevehu:
@handcraftedbits I don't see any problem with the update; however, I want to test on the Java 6 build on your repo to ensure that it is backward compatible. We still have a lot of users stuck with early version of Andriod and we have to support them. BTW, are you building the library on your own? What is your use case to support Java 11 build? Thanks a lot for your help.

handcraftedbits:
Yes, I'm building it for my own standalone project that uses JDK9+.  Thanks!

handcraftedbits:
FYI, I had one more change.  I bumped `slf4j-api` up to `1.7.30` because it looks like they did indeed add a fixed automatic name for their module in that version.  Also, I made the Jackson module requirement transitive.

stevehu:
https://github.com/networknt/json-schema-validator/issues/338

PR #328, opened 2020-10-03
Add '$comment' as a NonValidationKeyword for v7 and v2019 drafts

kmalski:
I am not sure I am able to  test this change. I did not find out any idea how to do it 😄 

You can find the reference to the schema here:
https://json-schema.org/understanding-json-schema/reference/generic.html#comments
and here:
https://json-schema.org/draft-07/json-schema-core.html#rfc.section.9


stevehu:
I don't think we need a test case for this as long as you don't see the warning message from the log. Thanks a lot for your help.

PR #341, opened 2020-11-02
Add contentMediaType, contentEncoding and examples as a NonValidationKeyword

jonnybbb:
`contentMediaType`, `contentEncoding` and `examples` should be treated as keywords according to the spec
https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.8.6 and https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.5


stevehu:
@jonnybbb Thanks a lof for your help.

PR #350, opened 2020-11-20
Add builder method that accepts iterable

wheelerlaw:
This lets me do something like this:

```java
JsonSchemaFactory factory = JsonSchemaFactory.builder(JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7))
            .uriFetcher(new JenkinsWorkspaceFileFetcher(), JenkinsWorkspaceFileFetcher.SUPPORTED_SCHEMES)
```

`JenkinsWorkspaceFileFetcher.SUPPORTED_SCHEMES` is a `Set<String>`, which can't be passed into varargs. But varargs can be passed into an `Iterable<String>` if you wrap it with `Arrays.asList(varargs)`, and it is cheap too since an `ArrayList<>` is just backed by an array.

### DIFF
--- a/src/main/java/com/networknt/schema/JsonSchemaFactory.java
+++ b/src/main/java/com/networknt/schema/JsonSchemaFactory.java
@@ -27,6 +27,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


stevehu:
@wheelerlaw It makes perfect sense. Thanks a lot for your help.

PR #352, opened 2020-11-24
YAML source location handling (#340)

### LINKED ISSUE #340: Suggestions on post validation line & column information?
One aspect we are really battling with.

How to log line/column information to a user for a YAML document that passes schema validation, but then fails semantic checks on the content of the JSON node tree?

Maybe we're missing something obvious here - but so far we have a multi step process:

1. Loading the YAML file/stream into JSON and then passing the loaded JSON tree to the factory for validation
2. Application code for processing the validated Node tree

After step 1 we have effectively lost the original source file information, so if we detect any semantic errors we can't easily report them back to the user with line/col information against the original file.

I doubt this is feasible, but I'm wondering if there couldn't be an option to include meta information in the processed JSON Node tree such as the original source positions of various nodes, which would then allow such error reporting.

ascertrobw:
Example describing how to retain YAML source file location information in processed and validated JSON tree

Fixes #340

### DIFF
--- a/doc/yaml-line-numbers.md
+++ b/doc/yaml-line-numbers.md
@@ -6,7 +6,7 @@ A great feature of json-schema-validator is it's ability to validate YAML docume
 
 ### Scenario 1, solution part 1 - capturing line details during initial parsing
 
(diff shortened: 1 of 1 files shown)


stevehu:
@ascertrobw Thanks for your help.

Issue #209, opened 2019-11-11
failFast fails too early in case of AnyOf validator


khiftikhar:
I have found the problem in failFast implementation. Consider this small change in the product schema

```
{
  "type": "object",
  "$schema": "http://json-schema.org/draft-04/schema#",
  "description": "A product from Acme's catalog",
  "title": "Product",
  "properties": {
    "id": {
      "type": "integer",
      "description": "The unique identifier for a product"
    },
    "name": {
      "type": "string",
      "maxLength": 200,
      "minLength": 10,
      "description": "Name of the product"
    },
    "price": {
      "anyOf": [
        {
          "type": "number",
          "minimum": 0,
          "exclusiveMinimum": true
        },
        {
          "type": "string",
          "minLength": 0
        }
      ]
    }
  },
  "required": [
    "id",
    "name",
    "price"
  ]
}
```
Now if the json object is like this, 
```
{
    "id": 1,
    "name": "A valid product name",
    "price": "10.99"
}
```
Now if failFast is set true, this schema will not validate. The reason behind this is that AnyOf collect errors in a loop, and there it calls `buildValidationMessage` which throws exception at once since failFast is set to true. However AnyOf validator is not finished yet, i.e. here `price` is not number but its a valid string, but we never let AnyOf validator to complete its validation completely and give its verdict. Therefore, we need not to call `buildValidationMessage`. 

Instead, AnyOf should throw `JsonSchemaException` after it is finished with its loop. 

Or @stevehu  you can guide me how should AnyOf should react in a more suitable way.

stevehu:
@khiftikhar AllOf, AnyOf, and OneOf work in a similar way with a context passed to the subschema. There are other behavior changes with other validators and that is why we have introduced the context. Take a look if we can figure out if the parent is AnyOf when deciding if throw the exception. @ddobrin As you are working on these aggregated validators, do you have any recommendation?

ddobrin:
I am out of the office until Wednesday and can look into it afterwards.

The delta between OpenAPI and JSON schemas will have to be bridged via the context, I would assume

TDog92:
if/then/else- condition also doesn´t work with failFast. When if-condition is not valid it fails before checking then-condition.

stevehu:
@TDog92 Thanks for capturing it. We definitely need to revisit the fail-fast implementation.

PR #372, opened 2021-01-29
Fix for issue-366

Krishna-capone:
The PR fixes the following:

For FailFast:

Now the OneOf Validator waits till all the subschemas are validated before throwing the exception instead of throwing the exception if the first subschema is invalid. I followed your suggestion for #209 (Looking to see if the parent schema is OneOf when deciding to throw the exception in BaseJsonValidator) and implemented it for OneOf.

If one of the subschema is valid and the other is not it does not throw exception which the current oneOf Validator does.

When all the subschemas are invalid it throws the following exception (this is an example for the test case provided):

com.networknt.schema.JsonSchemaException: $: should be valid to one and only one of the schemas but both schemas {{"type":"integer"}{"minimum":2}} are valid.

This has the detail about the subschemas. This is needed as there may be multiple oneOf conditions which are invalid in a schema and it would be difficult for the user to know which oneOf condition failed.

when neither of the subschemas are valid it throws the following exception:
com.networknt.schema.JsonSchemaException: [$: number found, integer expected, $: must have a minimum value of 2]

This has the detail about the subschemas. This detail is needed as there may be multiple oneOf conditions which are invalid in a schema and it would be difficult for the user to know which oneOf condition failed.

For Failslow:

when both of the subschemas are valid the following msg is sent:
$: should be valid to one and only one of the schemas but both schemas {{"type":"integer"}{"minimum":2}} are valid

when neither of the subschemas are valid the following messages are sent:
$: number found, integer expected
$: must have a minimum value of 2


stevehu:
@prubdeploy Could you please review the changes to ensure that your use case won't be impacted. It looks good to me in my initial review as all existing test cases are passed. Thanks.

guillomep:
I have the exact same case, where I want to retrieve "children" errors when the OneOf fails.

stevehu:
@guillomep It is nice to know this feature is welcomed. Could you please help to review the changes? Just want to make sure it covers your use case. Thanks.

src/main/java/com/networknt/schema/BaseJsonValidator.java
```diff
@@ -149,18 +149,22 @@ protected String getNodeFieldType() {
...
+    public Set<ValidationMessage> walk(JsonNode node, JsonNode rootNode, String at, boolean shouldValidateSchema) {
+        Set<ValidationMessage> validationMessages = new LinkedHashSet<ValidationMessage>();
+        if (shouldValidateSchema) {
+            validationMessages = validate(node, rootNode, at);
+        }
+        return validationMessages;
+    }
+
+    protected boolean isPartOfMultipleType(){
+        return parentSchema.schemaPath.equals(ValidatorTypeCode.ONE_OF.getValue());
```
guillomep, line 167: `ALL_OF` and `ANY_OF` are multiple types too, it should return true in these cases.
Or change method name to avoid confusion.

src/main/java/com/networknt/schema/OneOfValidator.java
```diff
@@ -174,26 +176,40 @@ public OneOfValidator(String schemaPath, JsonNode schemaNode, JsonSchema parentS
...
+            errors.add(message);
+        }
+        // ensure there is always an "OneOf" error reported if number of valid schemas is not equal to 1.
+        else if (numberOfValidSchema < 1) {
+            if (!childErrors.isEmpty()) {
+                errors.addAll(childErrors);
+            }
+            if(failFast){
+                throw new JsonSchemaException(errors.toString());
+            }
```
guillomep, line 197: Missing some spaces between "if" and parathesis

src/main/java/com/networknt/schema/OneOfValidator.java
```diff
@@ -209,4 +225,17 @@ public OneOfValidator(String schemaPath, JsonNode schemaNode, JsonSchema parentS
         return validationMessages;
     }
 
+    private ValidationMessage getBothSchemasValidErrorMsg(String at){
```
guillomep, line 228: What if we have 3 or more valid schemas ?

src/main/java/com/networknt/schema/BaseJsonValidator.java
guillomep: In this case maybe we have to do the same for ALL_OF and ANY_OF validator

prubdeploy:
> @prubdeploy Could you please review the changes to ensure that your use case won't be impacted. It looks good to me in my initial review as all existing test cases are passed. Thanks.
@stevehu The changes look good to me and doesn't impact my case and infact cover more scenarios as OneOf is considered.

src/main/java/com/networknt/schema/BaseJsonValidator.java
Krishna-capone: Thank you @guillomep. I thought I would fix ALL_OF and ANY_OF later. But for now i have renamed the method as you had suggested.

src/main/java/com/networknt/schema/OneOfValidator.java
Krishna-capone: Thank you @guillomep. I did not name the method properly. Thanks for catching it. I have changed the method name properly.

src/main/java/com/networknt/schema/OneOfValidator.java
Krishna-capone: Thank you @guillomep. I have added spaces as you suggested.

stevehu (approved):

stevehu:
Thanks a lot for your help. I have released a new version. 

https://github.com/networknt/json-schema-validator/releases/tag/1.0.48

guillomep:
Thank you !
Do you know when it will be available on Maven Central ?

stevehu:
Sonatype is down today. It will take some time for it to recovery. 

https://status.maven.org/?utm_source=embed

stevehu:
Sonatype is up and running. The release should be available from the maven central. Enjoy!!!

Issue #375, opened 2021-02-12
propertyNames not returning values used for validation


Eivyses:
Hey, I found an small bug with `propertyNames`. They do not return actual value used for validation. Example:
schema:
```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema#",
  "title": "test",
  "description": "asdasdint",
  "type": "object",
  "properties": {
    "fields": {
      "type": "object",
      "propertyNames": {
        "pattern": "^[a-zA-Z]+$",
        "maxLength": 5
      },
      "additionalProperties": {
        "type": "object",
        "properties": {
          "action": {
            "type": "string"
          }
        },
        "additionalProperties": false,
        "required": [
          "action"
        ]
      }
    }
  }
}
```

json:
```
{
  "fields": {
    "longName123": {
      "action": "doStuff"
    }
  }
}
```

Expected error:
```
Property name $.fields.longName123 is not valid for validation: maxLength 5"
```

Actual error:
```
Property name $.fields.longName123 is not valid for validation: maxLength"
```

Same applies to `pattern` and `minLength`.

Solution would be  to add second argument to error formatting:
https://github.com/networknt/json-schema-validator/blob/ff0a6fd22a8986e6343f0577d23569cac0414b61/src/main/java/com/networknt/schema/PropertyNamesValidator.java#L56-L57
```
errors.add(buildValidationMessage(at + "." + pname, "maxLength", entry.getValue().getSchemaNode().intValue())); // move to new variable, cast to String
```

stevehu:
@Eivyses Thanks a lot for pointing this out and providing the solution. Would you be able to submit a PR to get it fixed? I would rather see the credit goes to the real contributor. Thanks.

Eivyses:
@stevehu Thanks for quick reply, here's the PR: #376

PR #376, opened 2021-02-12
PropertyNames to return validator value on error (#375)

Eivyses:
* Fix for issue-375


stevehu (approved):
Nice PR with test cases!

Issue #383, opened 2021-03-13
Nested oneOf gives incorrect validation error under very specific circumstances


JonasProgrammer:
Hi,

first of all thank you for the effort to provide a Jackson-compatible JSONSchema validator.

I have recently encountered an error where both IntelliJ's integrated JSONSchema validator and an online validator accept a given input, but `json-schema-validator` rejects it.
The issue is related to state handling within `OneOfValidator`, but I was only able to trigger it by reducing my specific inputs -- just nesting `oneOf` did not do the job.

At its core this seems to be related to the parent still holding a reference to the once-pointed-to `validatorState`, even after `.remove()` is called on it and `nodeMatches` is false for that instance.

The following was the most reduced I could come up with that still triggers the issue; I'm aware that it is still verbose, so sorry for that.

```json
{
  "validation": {
    "all": [
      {
        "notEmpty": {
          "field": "test"
        }
      }
    ]
  }
}

```

```json
{
  "type": "object",
  "properties": {
    "validation": {
      "$ref": "#/definitions/predicateOrNull"
    }
  },
  "definitions": {
    "fieldRef": {
      "type": "object",
      "properties": {
        "field": {
          "type": "string"
        },
        "set": {
          "type": "string"
        }
      },
      "required": [
        "field"
      ]
    },
    "patternPredicate": {
      "type": "object",
      "oneOf": [
        {
          "properties": {
            "notEmpty": {
              "$ref": "#/definitions/fieldRef"
            }
          },
          "required": [
            "notEmpty"
          ]
        },
        {
          "properties": {
            "notBlank": {
              "$ref": "#/definitions/fieldRef"
            }
          },
          "required": [
            "notBlank"
          ]
        }
      ]
    },
    "allPredicate": {
      "type": "object",
      "properties": {
        "all": {
          "type": "array",
          "minItems": 1,
          "items": {
            "$ref": "#/definitions/predicate"
          }
        }
      },
      "required": [
        "all"
      ]
    },
    "predicate": {
      "type": "object",
      "oneOf": [
        {
          "$ref": "#/definitions/patternPredicate"
        },
        {
          "$ref": "#/definitions/allPredicate"
        }
      ]
    },
    "predicateOrNull": {
      "oneOf": [
        {
          "$ref": "#/definitions/predicate"
        },
        {
          "type": "null"
        }
      ]
    }
  }
}
```

JonasProgrammer:
Fixed in #384

PR #384, opened 2021-03-13
Test case and potential fix for #383

JonasProgrammer:
- signal nodeMatch to parent handler in OneOfValidator


stevehu:
@JonasProgrammer Thanks a lot for the contribution.

Issue #35, opened 2017-05-30
Add discriminator support


slinkydeveloper:
Hi,
I'm actually coding tools for OpenAPI 3 using your json schema validator. I know that it's a field out of JSON Schema Spec, but what do you think about implementing `discriminator` field of OpenAPI 3 spec? For more info: https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#fixed-fields-20

Thank you

stevehu:
@slinkydeveloper I have read the spec and it looks like easy to implement. We are in a process to refactor the library in order to support different versions of draft spec from 4 to 6 as well as customized validators. I will keep this open until the architecture change is done. I am a user of OpenAPI spec and I have a [framework](https://github.com/networknt/light-rest-4j) to generate the code and utilize the spec to do the validation and jwt scope verification at runtime. It is my best interest to support you:)

slinkydeveloper:
That's awesome thank you! I'm actually developing as gsoc student the support to openapi 3 inside vertx-web with runtime parameter validation and other stuff. If you want, take a look: https://github.com/slinkydeveloper/vertx-web/tree/designdriven/vertx-web/src/main/java/io/vertx/ext/web/validation

stevehu:
closed with most features implemented.

PR #390, opened 2021-03-23
Add discriminator support for #35

FWiesner:
Attempt to implement #35 for `anyOf` and `allOf`

### DIFF
--- a/src/main/java/com/networknt/schema/AnyOfValidator.java
+++ b/src/main/java/com/networknt/schema/AnyOfValidator.java
@@ -24,40 +24,79 @@ import java.util.*;
 
 public class AnyOfValidator extends BaseJsonValidator implements JsonValidator {
     private static final Logger logger = LoggerFactory.getLogger(RequiredValidator.class);
+    private static final String REMARK = "Remaining validation messages report why candidate schemas didn't match";
+    private static final String DISCRIMINATOR_REMARK = "and the discriminator-selected candidate schema didn't pass validation";
 
-    private List<JsonSchema> schemas = new ArrayList<JsonSchema>();
+    private final List<JsonSchema> schemas = new ArrayList<JsonSchema>();
+    private final ValidationContext validationContext;
+    private final ValidationContext.DiscriminatorContext discriminatorContext;
 
     public AnyOfValidator(String schemaPath, JsonNode schemaNode, JsonSchema parentSchema, ValidationContext validationContext) {
         super(schemaPath, schemaNode, parentSchema, ValidatorTypeCode.ANY_OF, validationContext);
+        this.validationContext = validationContext;
         int size = schemaNode.size();
         for (int i = 0; i < size; i++) {
-            schemas.add(new JsonSchema(validationContext, getValidatorType().getValue(), parentSchema.getCurrentUri(), schemaNode.get(i), parentSchema));
+            schemas.add(new JsonSchema(validationContext,
+                                       getValidatorType().getValue(),
+                                       parentSchema.getCurrentUri(),
+                                       schemaNode.get(i),
+                                       parentSchema));
+        }
+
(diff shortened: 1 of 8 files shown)


stevehu:
@FWiesner I looked into the implementation and haven't gone through the details on the OneOfValidator for all the logic yet. One thing that I want to point out is we might not need to create a new spec version OPENAPI_V3_0_3. The OpenAPI validator can use either V7 or V201909 for the validation. Is there any reason you want to create a new spec version? Thanks.

FWiesner:
well, this was just an idea. I've converted to the feature flag approach
you mentioned yesterday and will push a new commit with it today

Am Mi., 24. März 2021 um 02:17 Uhr schrieb Steve Hu <
***@***.***>:

src/main/java/com/networknt/schema/AnyOfValidator.java
```diff
@@ -24,40 +24,79 @@
...
+                } else if (config.isOpenAPI3StyleDiscriminators()) {
+                    if (discriminatorContext.isDiscriminatorMatchFound()) {
+                        if (!errors.isEmpty()) {
+                            errors.add(buildValidationMessage(at, DISCRIMINATOR_REMARK));
+                        }
+                        return errors;
+                    }
+                }
+                allErrors.addAll(errors);
+//                allErrors.add(renderMissingMatchValidationMessage(at));
```
FWiesner, line 86: @stevehu I wonder whether we can comment this line in - see the log renderer at the end of the file. At the moment failed `anyOf` validation doesn't give you an indication that no schema matched

stevehu: We have an issue opened to add more context information to the error messages to ensure they are understandable. If I am correct, the commented line will add an extra error message with more detailed info. Do you think it would be better to change the `errors.add(buildValidationMessage(at, DISCRIMINATOR_REMARK));` with the context info?

FWiesner: that code line will specifically log an additional line saying that none of the candidates matched. Richer context is certainly nice and might make my commented line obsolete. The question just is when that will come and whether then we could simply remove the line again.

Btw I've commented the line as it has impact on test conditions on bunch of tests

FWiesner:
the implementation has two limitations:
1.) I have not yet implemented a custom keyword to validate the discriminator
2.) the discriminator spec specifies that inline schemas get ignored. So,
```yml
Schema:
  properties:
    foo:
      type: string
  required: ["foo"]
  allOf:
    - $ref: ...
    - type: object
       properties:
         bar:
           type: string
       required: ["bar"]
```
should ignore `foo`, but enforce `bar`. Right now `foo` is not ignored

src/main/java/com/networknt/schema/AnyOfValidator.java
stevehu: OK. Let's leave it this way for now. We need to resolve this issue in one shot for all validators.

stevehu:
@FWiesner I think the limitations are OK for now as there is still something in discussion from the spec team. Could you please write a small document on how to use the new feature. I am sure a lot of people will start using it after upgrading to the latest specification. Thanks.

FWiesner:
Sure. What format?

stevehu:
We have a doc folder and all files are in markdown format and there is a link from the README.md to each feature.

stevehu:
1.0.51 is released. Thanks for your help. 
https://github.com/networknt/json-schema-validator/releases/tag/1.0.51

FWiesner:
Thanks Steve. I’m on vacation until tomorrow. Will send a doc file Thursday

Steve Hu ***@***.***> schrieb am Di. 30. März 2021 um 17:46:

PR #391, opened 2021-03-24
override default EmailValidator, if set custom email format

whirosan:
This pull request is overwrite the EmailValidator if I set custom email format.
The EmailValidator applied in "format: email" doesn't meet the requirements of my project, and I would like to overwrite it.
But, constructor of JsonSchema is private, and inner class Builder is forced to use EmailValidator in the case of "format: email" in the build method.

### DIFF
--- a/doc/validators.md
+++ b/doc/validators.md
@@ -80,3 +80,27 @@ You can use GroovyKeyword like below:
   }
 }
 ````
+
+### Override Email/UUID/DateTime Validator
+
+In this library, if the format keyword is "email", "uuid", "date", "date-time", default validator provided by the library will be used.
+
+If you want to override this behaivor, do as belows.
+
+```
+public JsonSchemaFactory mySchemaFactory() {
+    // base on JsonMetaSchema.V201909 copy code below
+    String URI = "https://json-schema.org/draft/2019-09/schema";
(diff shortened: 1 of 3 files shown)


stevehu:
@whirosan What is your customized email format? If it is generic, do you think it can replace the default one in the library? When I am looking at the solution, I am thinking about if we can make the solution to be applicable to all the other formats? What do you think?

whirosan:
@stevehu 
I want to use custom Pattern Fomat, because I have to meet the application's special EMail validation requirements.
I think that the validation of the email address may be changed depending on the application's requirements.
On the other hand, regarding UUID, date, and date-time, I don't feel the need for customization at the moment, but in some cases I may want to customize it.
I modified it so that it can be overridden by UUID, date, and date-time, so please review it.

stevehu:
@whirosan Look good to me. Would you be able to add a test case and write a small document so that others can learn how to use it? Thanks.

whirosan:
@stevehu Thank you very much.
I've added a small test and documentation, so please review it.

stevehu:
1.0.51 is released. Thanks for your help. 
https://github.com/networknt/json-schema-validator/releases/tag/1.0.51

PR #394, opened 2021-04-01
Added documentation for #390

FWiesner:


stevehu:
@FWiesner Thanks a lot for the detailed document.

Issue #396, opened 2021-04-11
Properly support propertyNames


JonasProgrammer:
The current implementation of `propertyNames` is completely separate from the remaining validation code, implementing special cases for stuff like `minLength`.

According to [the spec](https://json-schema.org/understanding-json-schema/reference/object.html#property-names), however, `propertyNames` should be treated like a normal json schema being applied to a text node of the property value.

This has lead to issues, such as #342 and #326.

I propose to just do what the spec says and apply the node content under `propertyNames` to each field name as if it were a `TextNode`.

PR #397, opened 2021-04-11
Proposed fix for #396: Implement propertyNames in terms full schema validation

JonasProgrammer:


stevehu:
@JonasProgrammer Nice fix!!! It was my mistake in the initial implementation due to the limited understanding of the document and the test cases. Thanks a lot for your help.

Issue #398, opened 2021-04-12
Two issues with OpenAPI 3 discriminators


FWiesner:
two issues we've uncovered while integrating the new discriminator feature into our solution.

1. referencing a schema with a discriminator, but without `anyOf` referencing it will lead to an `EmptyStackException`
2. not having found a matching candidate via the discriminator is not reliably reported.

I'll fix there two issues in the coming days

FWiesner:
solved in #399

PR #399, opened 2021-04-12
Fix for #398. 

FWiesner:
Also fixes the special situation of nested discriminators not working correctly when the parent schema has a discriminator, but is not referenced via anyOf/$ref

### DIFF
--- a/src/main/java/com/networknt/schema/AnyOfValidator.java
+++ b/src/main/java/com/networknt/schema/AnyOfValidator.java
@@ -83,7 +83,12 @@ public class AnyOfValidator extends BaseJsonValidator implements JsonValidator {
                     }
                 }
                 allErrors.addAll(errors);
-//                allErrors.add(renderMissingMatchValidationMessage(at));
+            }
+
+            if (config.isOpenAPI3StyleDiscriminators() && discriminatorContext.isActive()) {
+               final Set<ValidationMessage> errors = new HashSet<ValidationMessage>();
+               errors.add(buildValidationMessage(at, "based on the provided discriminator. No alternative could be chosen based on the discriminator property"));
+               return Collections.unmodifiableSet(errors);
             }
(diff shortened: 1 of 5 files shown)


src/main/java/com/networknt/schema/AnyOfValidator.java
```diff
@@ -83,7 +83,12 @@ public AnyOfValidator(String schemaPath, JsonNode schemaNode, JsonSchema parentS
                     }
                 }
                 allErrors.addAll(errors);
-//                allErrors.add(renderMissingMatchValidationMessage(at));
+            }
+
+            if (config.isOpenAPI3StyleDiscriminators() && discriminatorContext.isActive()) {
```
FWiesner, line 88: this fixes the case where validation passes, but the discriminator property value does not resolve to an alternative.

src/main/java/com/networknt/schema/JsonSchema.java
```diff
@@ -82,7 +82,7 @@ private JsonSchema(ValidationContext validationContext, String schemaPath, URI c
 
             if (config.isOpenAPI3StyleDiscriminators()) {
                 ObjectNode discriminator = (ObjectNode) schemaNode.get("discriminator");
-                if (null != discriminator) {
+                if (null != discriminator && null != validationContext.getCurrentDiscriminatorContext()) {
```
FWiesner, line 85: Needed for the case where the schema holds a discriminator, but it is not referenced from an `anyOf`, so is actually disabled. Why would you want to support this?

Say you have `Person` extended `Husband` and `Wife` with `Person` having a `spouse` parameter and a `discriminator`. Then you might want to use the `discriminator` only in a nested occurrence

FWiesner:
corrected the issue link. Should have been #398

stevehu (approved):

PR #403, opened 2021-05-14
Update to Java 11

chrswk:
Not sure what the stance of this project is in regards to newer Java versions. I've been using this library for a couple of years now and I would love to see it with a proper module info file. Feel free to change the module name if it's not appropriate and let me know, if there are any changes I should make.

-Chris


stevehu:
@chrswk I would love to upgrade to Java 11; however, there are still a lot of users stuck on Java 8 and needs this library. It took us a lot of discussions when we upgrade from Java 6 to Java 8 and we cannot simply upgrade to Java 11 and leave some of our users behind. I am open to ideas to upgrade to Java 11 but want to have another branch/release for Java 8 users if possible.

chrswk:
@stevehu I completely understand. I'll try to figure something out using [multi release](https://maven.apache.org/plugins/maven-compiler-plugin/multirelease.html)/[module info](https://maven.apache.org/plugins/maven-compiler-plugin/examples/module-info.html).

SiemelNaran:
Can we use Java 16 for the tests, but keep Java 8 for main?  This will allow me to use text blocks https://openjdk.java.net/jeps/355

stevehu:
@SiemelNaran The current source level is java 8 because a lot of users are still using it and we cannot leave them out. If we use java 16 for test cases, some developers might complain about it. Let's stay on Java 8 for now and think about upgrading when most users are upgraded.

stevehu:
@TJC I think it is a good idea to upgrade the plugins as long as the upgraded version still working with JDK8. Could you please submit a PR? I will test it with my local VM with Java8 installed. Thanks.

pom.xml
```diff
@@ -59,8 +59,9 @@
         </repository>
     </distributionManagement>
     <properties>
-        <java.version>1.6</java.version>
-        <java.testversion>1.8</java.testversion>
+        <maven.compiler.release>11</maven.compiler.release>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
```
rosti-il, line 64: If the `maven.compiler.release` property is used the older `maven.compiler.source` and `maven.compiler.target` properties should be omitted.

pom.xml
```diff
@@ -237,20 +238,17 @@
...
-                            <source>8</source>
                         </configuration>
                     </execution>
                 </executions>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.6.1</version>
+                <version>3.8.1</version>
```
rosti-il, line 248: Why not the latest 3.10.1 version?

SiemelNaran: This PR is old, that's why it's an old version

stevehu: Due to many of our users are still using Java 8, we cannot migrate to Java 11  in the current form. We are open to all options to support both Java 8 and Java 11.

agentgt:
If this project upgrades to Java 11 I recommend Apache Commons lang3 get removed since all of its uses appears to be to use `StringUtils.isNotBlank`. Java 11 has `String.isBlank` builtin.

Issue #408, opened 2021-06-15
uri format regexp is validating invalid URI


vmaurin:
The regexp used by the "uri" format validation doesn't really validate that the string is a proper URI, i.e there is a chance passing a validated string to the java URI class will result into an exception

An example invalid URI that passes the validation is `https://gitlab.com/?args=invalid|value`

Should the regexp be more complex ? Should the validation be made with the java URI class (then catching the exception to validate) ?

vmaurin:
The regexp in question is this one https://github.com/networknt/json-schema-validator/blob/72fe835006931dc6b7b71f61c92ce48d5635b782/src/main/java/com/networknt/schema/JsonMetaSchema.java#L50

stevehu:
@vmaurin The regexp definitely has room to improve. The reason I didn't use the Java URI class to verify the URL is due to performance. Most users are using this library to validate the HTTP request and it needs to be as fast as possible. I would suggest improving the regexp to make it cover more validation scenarios. The other option is to provide both options but default to regexp and with a configuration flag switch to Java URL class. What do you think?

vmaurin:
I totally agree with the performance issue on relying on try/catch for validation, and also after a quick research, it sounds that URI java class is not perfect regarding validation either. My personal pick would be as you suggested, to enrich a bit the regexp. There are several example online, with this one as the most complete validation http://jmrware.com/articles/2009/uri_regexp/URI_regex.html but it might be also overkill. There are also some middle ground, like this one `^([a-z][a-z0-9+.-]+):(\/\/([^@]+@)?([a-z0-9.\-_~]+)(:\d+)?)?((?:[a-z0-9-._~]|%[a-f0-9]|[!$&'()*+,;=:@])+(?:\/(?:[a-z0-9-._~]|%[a-f0-9]|[!$&'()*+,;=:@])*)*|(?:\/(?:[a-z0-9-._~]|%[a-f0-9]|[!$&'()*+,;=:@])+)*)?(\?(?:[a-z0-9-._~]|%[a-f0-9]|[!$&'()*+,;=:@]|[/?])+)?(\#(?:[a-z0-9-._~]|%[a-f0-9]|[!$&'()*+,;=:@]|[/?])+)?$` that could be implemented, I just don't know what should be the position of json-schema-validator on this regards

stevehu:
It looks good to me. Would you like to submit a PR to replace the old one? Thanks.

vmaurin:
[...]

PR #409, opened 2021-06-16
Use a more restrictive regex for URI validation

vmaurin:
The json schema uri format should follow rfc3986 for uri validation. I
have replaced the current regex with a more restrictive one in order to
have a stronger validation according the RFC

refs https://github.com/networknt/json-schema-validator/issues/408


stevehu:
A new release is out. https://github.com/networknt/json-schema-validator/releases/tag/1.0.54

PR #410, opened 2021-06-21
Added JsonValidator#preloadJsonSchema()

### LINKED ISSUE #406: Behavior change of $ref resolution
In our system we support some well known URNs that our clients can use in their schema definitions, using the `$ref` keyword, for example:

```
{
    "properties": {
        "someField": {
            "$ref": "urn:my-company:my-custom-type"
        }
    }
}
```

We use a custom `URNFactory` and works great.

But we have observed the way of the URN resolution has changed as of release `1.0.49`. Previously, if a schema contained an unknown URN the call to `JsonSchemaFactory.getSchema(...)` failed with a `JsonSchemaException`. Now it does not fail in that moment but later, when calling to `JsonSchema.validate(...)`. I suspect this change could be due to https://github.com/networknt/json-schema-validator/pull/377.

Is is possible to recover the old behavior and resolve the `$ref` keywords when getting the schema? that way we would ensure the schema instantiated is valid, or at least would it be possible to provide a way to load the `$ref` schemas eagerly after getting the schema instance?

FWiesner:
Hi @stevehu,

to fix #406, here's my proposal. I've also done a few code clean-ups, especially when member `JsonSchema` instances were not final. Please refer to `Issue406Test` for the behaviour of the change.

Florian

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -69,6 +69,10 @@ class Sample extends BaseJsonSchemaValidatorTest {
         JsonNode schemaNode = getJsonNodeFromStringContent(
             "{\"$schema\": \"http://json-schema.org/draft-06/schema#\", \"properties\": { \"id\": {\"type\": \"number\"}}}");
         JsonSchema schema = getJsonSchemaFromJsonNodeAutomaticVersion(schemaNode);
+        
+        schema.initializeValidators(); // by default all schemas are loaded lazily. You can load them eagerly via
+                                       // initializeValidators() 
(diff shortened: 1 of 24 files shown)


FWiesner:
@stevehu as requested, I've updated the quickstart

stevehu:
@FWiesner Thanks a lot for your help.

stevehu:
A new release is out. https://github.com/networknt/json-schema-validator/releases/tag/1.0.54

Issue #411, opened 2021-06-23
With #409 http://json-schema.org/draft-07/schema# is not accepted as valid schema anymore


FWiesner:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "...": "..."
}
```
now leads to

```
$.$schema: does not match the uri pattern ^([A-Za-z][A-Za-z0-9+.-]+):(\/\/([^@]+@)?([A-Za-z0-9.\\-_~]+)(:\\d+)?)?((?:[A-Za-z0-9-._~]|%[A-Fa-f0-9]|[!$&'()*+,;=:@])+(?:\/(?:[A-Za-z0-9-._~]|%[A-Fa-f0-9]|[!$&'()*+,;=:@])*)*|(?:\/(?:[A-Za-z0-9-._~]|%[A-Fa-f0-9]|[!$&'()*+,;=:@])+)*)?(\\?(?:[A-Za-z0-9-._~]|%[A-Fa-f0-9]|[!$&'()*+,;=:@]|[/?])+)?(\\#(?:[A-Za-z0-9-._~]|%[A-Fa-f0-9]|[!$&'()*+,;=:@]|[/?])+)?$
```

FWiesner:
@vmaurin could you please have a look?

vmaurin:
Sure. When did you get the issue ? I run my own test with json schema 7 and I didn't got the problem. Could you provide a java code snippet leading to this error ?

FWiesner:
load the JSON Schema Draft 7 meta schema from json-schema.org and then validate itself against it

vmaurin:
Ok according to the RFC, empty fragment is allowed https://datatracker.ietf.org/doc/html/rfc3986/#appendix-A , the regexp I used is bugged in this regard. I will open a PR for the fix

FWiesner:
thanks!

vmaurin:
Empty query string also might be allowed, I will try too

vmaurin:
@FWiesner The PR I have opened should fix the issue (and also probable one with empty query string). Could you test with it ?

stevehu:
@vmaurin @FWiesner Thanks a lot for your effort. I have merged the PR. Once you guys confirm it is working I will release another version.

FWiesner:
@stevehu @vmaurin we currently test it

FWiesner:
@stevehu please release - it is working now

vmaurin:
[...]

PR #412, opened 2021-06-23
Fix empty fragment and query string being invalid

vmaurin:
According to RFC 3986, URI fragment could be empty. The previous regexp
was only validating non empty fragment, resulting false positive URI
invalidation

refs https://github.com/networknt/json-schema-validator/issues/411


stevehu:
@vmaurin Thanks a lot for the quick fix. A new version is released. https://github.com/networknt/json-schema-validator/releases/tag/1.0.55

Issue #413, opened 2021-06-24
Stackoverflow on format `uri` validation


vmaurin:
I ran into an other problem with the regex I commit for URI validation. It is using a lot of `|` that are quite intensive regarding the text evaluation. As java Pattern class seems implemented with recursion, it can quickly fill the stack and one might run into Stackoverflow errors while evaluating an URI relatively large (thousand of chars)

@stevehu @FWiesner You are probably interested on this issue too. I will try to quickly rework again the regular expression, to have something that might be more tolerant, but with lower impact on the call stack (avoiding all `|`)

stevehu:
@vmaurin Thanks for the findings. This is the way we make the library more useful for all users. Iterative refinement is the key to make this library grow.

vmaurin:
@stevehu I have opened the PR, I have left as comment the process of though to rework the regexp, not sure if it is useful, let me know on the PR

ruseinov:
https://github.com/jonckvanderkogel/java-tail-recursion may be of use

stevehu:
@ruseinov Thanks for the link. Interesting example.

vmaurin:
Fixed in https://github.com/networknt/json-schema-validator/releases/tag/1.0.56

PR #414, opened 2021-06-25
Simplify the uri format validation regexp

vmaurin:
The previous regexp was relying a lot on `|` leading to stackoverflow
errors for long URIs (as Java Pattern class is using recursion a lot). I
have reworked the regexp from the RFC directly, taking some shortcut to
mostly validate what are the allowed or not allowed char in the
different part of the URI

refs https://github.com/networknt/json-schema-validator/issues/413


stevehu:
Thanks a lot for the effort to optimize it. Let me know if you need a quick release.

stevehu:
A new release is published. Thanks. 

https://github.com/networknt/json-schema-validator/releases/tag/1.0.56

vmaurin:
Thank you @stevehu

Issue #79, opened 2018-04-03
Warning when validating a schema


rhwood:
I am using this library to validate a service's JSON outputs and one of the outputs are the JSON schema (against which everything is validated) themselves.

When validating the schema, I get this warning:
```
WARN  - Unknown keyword exclusiveMinimum - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword [main] com.networknt.schema.JsonMetaSchema.newValidator()
```
The only place I see exclusiveMinimum is in the JSON Schema (Draft 4) itself. What is the best way to avoid this warning, since I am not using it anywhere in my own schemas?

stevehu:
@rhwood That warning message can be ignored. We are loading the JSON schema draft v4 for validation and that particular keyword is not supported yet. We are planning to refactor the library to support both v4 and v6 in the near future and this issue will be resolved at the same time. Let's keep this open until it is resolved. Thanks for bringing it up.

alisrael:
Has this been resolved yet?

stevehu:
@alisrael I just mark the issue with `help wanted` and hope someone from the community can pick it up and get it resolved. Thanks for bringing it up.

markgitset:
Can the warning be suppressed?

stevehu:
If exclusiveMinimum is the keyword in draft 4 only, we can map this keyword to TrueValidator for v6, v7 and 2019-09. Let's wait for the test result on this issue and make the same change if it works. 

https://github.com/networknt/json-schema-validator/issues/333

stevehu:
Since this issue was raised, we have added a lot more NonValidationKeyword to each version of specifications. Here is an example. https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/JsonMetaSchema.java#L172

I am wondering if you guys could try it again and add your warned keyword to the list. Just be sure that don't add anything that is a validation keyword but not implemented yet. Thanks.

bomzheg:
still not fixed?

2021-11-30 18:09:38.189 [main] WARN  com.networknt.schema.JsonMetaSchema - Unknown keyword $defs - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword
2021-11-30 18:09:39.192 [main] WARN  com.networknt.schema.JsonMetaSchema - Could not load validator maxLength

stevehu:
[...]

PR #418, opened 2021-07-02
201909 false flag keywords "additonalItems", "then"

pgalbraith:


stevehu:
@pgalbraith Thanks a lot for your help. It makes perfect sense to add `then` to the list as we only check the `if` keyword in our implementation. However, I think we should implement additionalItems if possible. What do you think? Compare with items, it is not popular. But if it is present in your schema, I think we need to respect the validation rule defined here. 

https://json-schema.org/draft/2019-09/json-schema-core.html#additionalItems

pgalbraith:
Hmmm the details of the spec are beyond me, I'm afraid.  My goal is just to prevent the non-keyword warning message from showing (i.e. #79) when I'm validating a simple json doc.  My schema references the 2019-09 meta schema, and I'm getting these warning messages when validating documents:

```
WARN com.networknt.schema.JsonMetaSchema - Unknown keyword then - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword
WARN com.networknt.schema.JsonMetaSchema - Unknown keyword additionalItems - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword

```

stevehu:
@pgalbraith Let's first prevent the message shown up. The additionalItems keyword is defined in the spec, but I don't think anybody is using it. We can implement it when someone needs it.

Issue #425, opened 2021-07-09
Cannot distinguish the "TextNode" and the "ArrayNode" with single value for oneOf


rongyj:
When schema has "oneOf" as below:
```
  "values": {
            "oneOf": [
              {
                "type": "string",
                "nullable": true
              },
              {
                "type": "array",
                "nullable": true,
                "items": {
                  "type": "string"
                }
              }
            ]
          },
 ```
When input test data is :
```
values: "test"
```
It has a validation error:
```
should be valid to one and only one of the schemas but more than one schemas {{"type":"string","nullable":true}{"type":"array","nullable":true,"items":{"type":"string"}}} are valid
```
when input data is :
```
values : ["test"]
```
It is fine.

rongyj:
Create a PR https://github.com/networknt/json-schema-validator/pull/432 which include the fixes for issue428 too.

PR #431, opened 2021-08-11
Issue428

rongyj:
This PR including:
1. Test data and unit test case for Issue428
2. Handle nullable for the "oneOf" and the child type nodes
3. When there is childErrors, skip the check for the next validation against a different child type node

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ dist/
 .settings
 .metadata/
 *.iml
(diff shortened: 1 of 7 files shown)


stevehu (approved):

stevehu:
@rongyj Thanks a lot for your help. I have merged this PR and the #425 needs to resolve the conflicts.

PR #432, opened 2021-08-11
Issue425

rongyj:
This PR is on the top of PR for issue428. 
It skip the validation when the test node is oneOf the defined type and schemaType is not match to the nodeType.


rongyj:
I also did small refactoring to move the "equalsToSchemaType" in TypeValidator to "JsonNodeUtil" for reuse.

PR #435, opened 2021-08-17
Added exampleSetFlag to nonValidationKeyword

ShubhamRwt:
Issue:

When using vertx, there is a keyword `exampleSetFlag` present in swagger which is transitive dependency of vertx . So when we run the verticle we get a warning like this :- 
`<onMetaSchema:338> [oop-thread-1]  Unknown keyword exampleSetFlag - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword`

PR purpose :- 

It adds the variable `exampleSetFlag` to the nonValidationKeyword in V4 since the OpenApiv3 use the Json schema v4 --> `    private static final JsonMetaSchema META_SCHEMA = JsonMetaSchema.builder(JsonMetaSchema.getV4().getUri(), JsonMetaSchema.getV4()).addKeyword(new NonValidationKeyword("example")).build();`

### DIFF
--- a/src/main/java/com/networknt/schema/JsonMetaSchema.java
+++ b/src/main/java/com/networknt/schema/JsonMetaSchema.java
@@ -117,7 +117,8 @@ public class JsonMetaSchema {
                             new NonValidationKeyword("title"),
                             new NonValidationKeyword("description"),
                             new NonValidationKeyword("default"),
-                            new NonValidationKeyword("definitions")
+                            new NonValidationKeyword("definitions"),
+                            new NonValidationKeyword("exampleSetFlag")
                     ))
                     .build();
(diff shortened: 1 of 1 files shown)


stevehu:
This is a temp solution to compensating a bug from the swagger. And it might be removed in the future as the exampleSetFlag is not a JSON Schema keyword.

```
I did a search and seems this issue has been resolve by the swagger team already. swagger-api/swagger-core#3702
Also, it is only a warning message in the log, it this a big concern for your team? I am OK to accept a PR as a temporary workaround for the Swagger bug and you should change the version you are using not the v4 I mentioned in the example above.
```

PR #438, opened 2021-08-20
"Adding custom message support"

adilath18:
We can create custom message, by adding another property in schema i.e **message**  for e.g.

If our schema is like this:

```
"schema": {
      "properties": {
        "foo": {
          "type": "integer"
        },
        "bar": {
          "type": "string"
        }
      }
    },
```

Then to provide custom message, we have to modify it to:

```
"schema": {
      "properties": {
        "foo": {
          "type": "integer"
        },
        "bar": {
          "type": "string"
        }
      },
    **"message": {
        "type" : "Invalid type"
      }**
    },
```

Since we are getting arguments in ValidationMessage, we can make this message more illustrative using dynamic arguments.
e.g.  **{0} invalid type** where , {0}  will be path


stevehu:
This is a smart way to implement the customized error message with the schema. Could you please add a small doc on the feature so that other people will benefit from your work? Thanks.

PR #443, opened 2021-09-06
$ref caching issue

prashanthjos:
In a scenario where we use the same `JsonSchemaFactory` to create multiple `JsonSchema` instances with each instance having its own config (which is instance of `SchemaValidatorsConfig`) the cached $ref's are picking up the initial config that was set by first call to `JsonSchemaFactory.getSchema`. This PR addresses the issue of refreshing the config every time we call `JsonSchemaFactory.getSchema`. As per my understanding we should be caching the JSONNode instance associated with a schema and not the config.

@stevehu can you please have a look at the PR and let me know your suggestions and approve it if we are good.


prashanthjos:
@stevehu can you please have a look at this PR.

prashanthjos:
@stevehu this review has been pending for a while. We have an important functionality that is waiting for this PR to be merged. Can you please take a look and let me know your comments.

stevehu (approved):

Issue #461, opened 2021-10-05
1.0.60 Breaks Nested Schema Types


tomdeering-wf:
When dependabot brought us from using json-schema-validator 1.0.59 to 1.0.60 in an internal project, many tests started failing. The root cause seems to be that json-schema-validator 1.0.60 expects that `"type"` should be an array rather than a simple string, even though a simple string [is allowed by the spec](http://json-schema.org/draft-07/schema#).

Here is a test case value reproduces the issue:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string"
    },
    "lastName": {
      "type": "string"
    },
    "address": {
      "type": "object",
      "properties": {
        "street": {
          "type": "string"
        },
        "city": {
          "type": "string"
        }
      },
      "required": [
        "street",
        "city"
      ]
    }
  },
  "required": [
    "firstName",
    "lastName",
    "address"
  ]
}
```

Validated using:

```java
final var config = new SchemaValidatorsConfig();
config.addKeywordWalkListener(ValidatorTypeCode.PROPERTIES.getValue(), new JsonSchemaWalkListener() {
      private final List<ParseResult.Property> path = Lists.newArrayList();

      @Override
      public WalkFlow onWalkStart(final WalkEvent walkEvent) {
        // This method is more confusing than it should be on first blush. The reason for this is
        // that technically properties can have the literal name "properties" and can also contain
        // periods which means we need complex logic as we can use simple string splitting or
        // jsonpath. We have to track the state of our recursive descent to strip out known prefixes
        // from our current path.

        // The first event is always the outer object properties.
        if (walkEvent.getAt().equalsIgnoreCase("$")) {
          path.add(new ParseResult.Property("properties", "$.properties"));
          return WalkFlow.CONTINUE;
        }

        final var previous = path.get(path.size() - 1);
        final var path = walkEvent.getAt();

        var name = StringUtils.removeStart(path, previous.getPath() + ".");
        final var property = new ParseResult.Property(name, walkEvent.getAt());

        properties.add(property);

        final var nextProperty = new ParseResult.Property(name, walkEvent.getAt() + ".properties");
        this.path.add(nextProperty);

        return WalkFlow.CONTINUE;
      }

      @Override
      public void onWalkEnd(final WalkEvent walkEvent,
                            final Set<ValidationMessage> validationMessages) {

        this.path.remove(path.size() - 1);
      }
    });
[...]

PR #462, opened 2021-10-06
fix for #461 

bartoszm:
there is also a bit simplified version https://github.com/networknt/json-schema-validator/pull/463 of the fix that is aligned with one-of implementation. however, I think this one is more accurate.


stevehu:
Thanks a lot for your help.

tomdeering-wf:
Thank you @bartoszm!

PR #463, opened 2021-10-06
fix for #461 simple implementation (aligned with one-of)

bartoszm:


bartoszm:
I will close it - I attached it only for demonstration purposes.

PR #465, opened 2021-10-10
Example of problems with $id resolution

bartoszm:
This test fails when in my opinion it should pass.


stevehu:
@bartoszm Could you please verify if the issue still exists with the latest version in the master branch? Thanks.

PR #472, opened 2021-10-18
Fix i18n doesn't work with locale CHINA

wyzfzu:
1. rename file jsv-messages_zh_cn.properties to jsv-messages_zh_CN.properties
2. add some test methods


stevehu:
@wyzfzu Thanks a lot for the fix. I have released 1.0.63 and the release note can be found at https://github.com/networknt/json-schema-validator/releases/tag/1.0.63

PR #473, opened 2021-10-20
Remove additional properties validation message when there are other …

jsu216:
…types of message in OneOfValidator


stevehu:
@jsu216 Would you like to add a test case to ensure that your changes won't be broken in the future update to this validator? This validator is very complicated, so the more test cases we have the more confident we have to make enhancements to it. Thanks.

PR #477, opened 2021-10-30
apply default in objects and arrays

SiemelNaran:
At https://json-schema.org/understanding-json-schema/reference/generic.html they say of the default keyword

> The default keyword specifies a default value. This value is not used to fill in missing values during the validation process. Non-validation tools such as documentation generators or form generators may use this value to give hints to users about how to use a value. However, default is typically used to express that if a value is missing, then the value is semantically the same as if the value was present with the default value. The value of default should validate against the schema in which it resides, but that isn’t required.

So before this PR, the 'default' keyword is ignored during both validation and walking.  But this PR allow the user to apply defaults during walking.  They still have to set a flag indicating that they want defaults to be applied (and they can apply defaults only if property is missing, or if it missing or is explicitly null).

If JSON blobs were saved to the database, and the JSON schema changed to add defaults, the db migration tool could use read the JSON blobs and walk them against the schema again to apply defaults.

### DIFF
--- a/src/main/java/com/networknt/schema/JsonSchema.java
+++ b/src/main/java/com/networknt/schema/JsonSchema.java
@@ -20,12 +20,14 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLDecoder;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -77,7 +79,8 @@ public class JsonSchema extends BaseJsonValidator {
     private JsonSchema(ValidationContext validationContext, String schemaPath, URI currentUri, JsonNode schemaNode,
                        JsonSchema parent, boolean suppressSubSchemaRetrieval) {
         super(schemaPath, schemaNode, parent, null, suppressSubSchemaRetrieval,
-              validationContext.getConfig() != null && validationContext.getConfig().isFailFast());
+              validationContext.getConfig() != null && validationContext.getConfig().isFailFast(),
+              validationContext.getConfig() != null ? validationContext.getConfig().getApplyDefaultsStrategy() : null);
         this.validationContext = validationContext;
         this.idKeyword = validationContext.getMetaSchema().getIdKeyword();
         this.currentUri = this.combineCurrentUriWithIds(currentUri, schemaNode);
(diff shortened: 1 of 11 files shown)


SiemelNaran:
mvn test results in 4 errors

```
[ERROR] Failures: 
[ERROR]   OpenAPI30JsonSchemaTest.testDiscriminatorMapping:120->runTestFile:109 expected: <false> but was: <true>
[ERROR]   V4JsonSchemaTest.testNotValidator:155->BaseSuiteJsonSchemaTest.runTestFile:116 expected: <false> but was: <true>
[ERROR]   V7JsonSchemaTest.testNotValidator:272->BaseSuiteJsonSchemaTest.runTestFile:116 expected: <false> but was: <true>
[ERROR] Errors: 
[ERROR]   CollectorContextTest.testCollectorGetAll:109 » NullPointer
[INFO] 
[ERROR] Tests run: 310, Failures: 3, Errors: 1, Skipped: 46
```

src/main/java/com/networknt/schema/JsonSchema.java
```diff
@@ -239,6 +239,20 @@ private boolean nodeContainsRef(String ref, JsonNode node) {
         return validators;
     }
 
+    /**
+     * A comparator that sorts validators, such such that 'properties' comes before 'required',
+     * so that we can apply default values before validating required.
+     */
+    private static Comparator<String> VALIDATOR_SORT = (lhs, rhs) -> {
```
SiemelNaran, line 246: This change also changes the order of validators, so the order of validation messages will change.  If anyone was expecting the old order, they will break. Fortunately mvn test passes, and my code uses Matchers.containsInAnyOrder.

stevehu: I don't think the order of the messages is important.

SiemelNaran: What about performance concerns?  Changing HashMap to TreeMap is changing O(1) to O(lg(N)).  Is there a performance test that can be run to test?

stevehu: The focus of this json-schema-validator is the small size of the JSON object and the number of the validators shouldn't that big. In most cases, we shouldn't worry about the performance. For the large-size JSON schemas, we might have an issue. I would suggest picking up one big JSON schema to test it with this [repo](https://github.com/networknt/json-schema-validator-perftest). If we see a performance issue, we can make it configurable to use either HashMap or TreeMap in the code. What do you think?

SiemelNaran: (1) Using json-schema-validator-perftest as is which is version 1.0.49 time is around 430ms

(2) Upgrading to version 1.0.63 by changing the version to
```
            <version>1.0.63</version>
```
and the time is now around 465ms.

(3) Using my code by adding

            <scope>system</scope>
            <systemPath>/Users/snaran/main/Software/json-schema-validator/target/json-schema-validator-1.0.63.jar</systemPath>

now time is around 430ms.  Maybe the TreeMap makes things faster.  Hash map has to hash the whole string, which is O(N) of the whole string, whereas TreeMap only has to compare chars till it finds the first difference, so in theory TreeMap could be faster sometimes.  But who knows, just guessing.

SiemelNaran:
Reminder about this review in case you missed it @prashanthjos .  @stevehu I did the performance tests and replied in the thread, and to my surprise found that the code is faster after my changes.

stevehu:
@SiemelNaran That is good news :) Let us know when you think the PR is completed so that we can start the review. Thanks a lot for the effort.

SiemelNaran:
> @SiemelNaran That is good news :) Let us know when you think the PR is completed so that we can start the review. Thanks a lot for the effort.

It is ready.

Issue #386, opened 2021-03-16
if-then does not respect enum values


0xShamil:
Hi. 

The `if-then` construct with the combination of `allOf` does not respect the `enum`s defined in the property value. The properties value as provided in the `if` construct is alone considered valid. 

[Considering the example from the conditionals page](https://json-schema.org/understanding-json-schema/reference/conditionals.html):
```json
{
  "type": "object",
  "properties": {
    "street_address": {
      "type": "string"
    },
    "country": {
      "enum": ["United States of America", "Canada", "Netherlands"]
    }
  },
  "allOf": [
    {
      "if": {
        "properties": { "country": { "const": "United States of America" } }
      },
      "then": {
        "properties": { "postal_code": { "pattern": "[0-9]{5}(-[0-9]{4})?" } }
      }
    },
    {
      "if": {
        "properties": { "country": { "const": "Canada" } }
      },
      "then": {
        "properties": { "postal_code": { "pattern": "[A-Z][0-9][A-Z] [0-9][A-Z][0-9]" } }
      }
    },
    {
      "if": {
        "properties": { "country": { "const": "Netherlands" } }
      },
      "then": {
        "properties": { "postal_code": { "pattern": "[0-9]{4} [A-Z]{2}" } }
      }
    }
  ]
}
```

The `JsonSchema` is constructed as follows:

```java
  JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
  SchemaValidatorsConfig config = new SchemaValidatorsConfig();
  config.setFailFast(true);
  InputStream stream = // read schema
  JsonSchema schema = factory.getSchema(stream, config);
```

If I take the first example
```
{
  "street_address": "1600 Pennsylvania Avenue NW",
  "country": "United States of America",
  "postal_code": "20500"
}
```
then I get the following error:

```
com.networknt.schema.JsonSchemaException: $.country: must be a constant value Canada
	at com.networknt.schema.BaseJsonValidator.buildValidationMessage(BaseJsonValidator.java:130)
	at com.networknt.schema.ConstValidator.validate(ConstValidator.java:44)
	at com.networknt.schema.JsonSchema.validate(JsonSchema.java:224)
	at com.networknt.schema.PropertiesValidator.validate(PropertiesValidator.java:75)
	at com.networknt.schema.JsonSchema.validate(JsonSchema.java:224)
	at com.networknt.schema.IfValidator.validate(IfValidator.java:54)
[...]

PR #485, opened 2021-11-22
Issue386: FailFast should not cause exception on if

gareth-robinson:
Fix for https://github.com/networknt/json-schema-validator/issues/386

When failFast is enabled the code that builds the ValidationMessages throws an exception rather than returning a Set.
In if/else/then blocks this could cause the validation to fail whenever an 'if' was false.
I've a basic change here to just wrap the if validation in a try/catch if that's appropriate?

Btw, when running the tests I found that Issue366Test wasn't working, as it was looking in the wrong location (but otherwise is not related to this change)

### DIFF
--- a/src/test/java/com/networknt/schema/Issue366FailFast.java
+++ b/src/test/java/com/networknt/schema/Issue366FailFast.java
@@ -35,7 +35,7 @@ public class Issue366FailFast {
 
     URI uri = getSchema();
 
-    InputStream in = getClass().getResourceAsStream("/draft7/issue366_schema.json");
+    InputStream in = getClass().getResourceAsStream("/schema/issue366_schema.json");
     JsonNode testCases = objectMapper.readValue(in, JsonNode.class);
     this.jsonSchema = schemaFactory.getSchema(uri, testCases,schemaValidatorsConfig);
   }
@@ -75,21 +75,21 @@ public class Issue366FailFast {
   @Test
   public void bothValid() throws Exception {
     String dataPath = "/data/issue366.json";
-    
+
     assertThrows(JsonSchemaException.class, () -> {
         InputStream dataInputStream = getClass().getResourceAsStream(dataPath);
         JsonNode node = getJsonNodeFromStreamContent(dataInputStream);
         List<JsonNode> testNodes = node.findValues("tests");
         JsonNode testNode = testNodes.get(0).get(2);
         JsonNode dataNode = testNode.get("data");
-        jsonSchema.validate(dataNode);        
+        jsonSchema.validate(dataNode);
     });
   }
 
   @Test
   public void neitherValid() throws Exception {
     String dataPath = "/data/issue366.json";
-    
+
     assertThrows(JsonSchemaException.class, () -> {
         InputStream dataInputStream = getClass().getResourceAsStream(dataPath);
         JsonNode node = getJsonNodeFromStreamContent(dataInputStream);
--- a/src/test/java/com/networknt/schema/Issue366FailSlowTest.java
+++ b/src/test/java/com/networknt/schema/Issue366FailSlowTest.java
@@ -1,7 +1,7 @@
 package com.networknt.schema;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -13,7 +13,7 @@ import java.util.Set;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-public class Issue366FailFast {
+public class Issue366FailSlowTest {
 
   @BeforeEach
   public void setup() throws IOException {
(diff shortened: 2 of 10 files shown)


src/test/resources/data/issue386.json
```diff
@@ -0,0 +1,5 @@
+{
```
SiemelNaran, line 1: Can we have all of the test cases in https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else

You could maybe make two arrays like so

```
{
  "passing": [
      { your example above },
      { other passing example from the link },
      ...
    ],
   "failing": [
       { "data": { a failing example from the link }, "expectedErrors":["error1", "error2"] },
       ...
}
```

src/test/java/com/networknt/schema/Issue386Test.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Test
+    public void dataIsValidFailFast() throws Exception {
+        String schemaPath = "/schema/issue386-v7.json";
+        String dataPath = "/data/issue386.json";
+        JsonSchema schema = getJsonSchemaFromPathV7(schemaPath, true);
+        JsonNode node = getJsonNodeFromPath(dataPath);
+        Set<ValidationMessage> errors = schema.validate(node);
+        Assertions.assertEquals(0, errors.size());
+    }
```
SiemelNaran, line 46: These tests look the same, except for failFast.  So consider using parametrized testing.  You can get pom.xml changes from https://github.com/networknt/json-schema-validator/pull/477/files.

Then the test is something like

```
@ParameterizedTest
@ValueSource(booleans = { true, false})
void dataIsValid(boolean failFast) {
}
```

src/test/java/com/networknt/schema/Issue366FailFast.java
```diff
@@ -35,7 +35,7 @@ private void setupSchema() throws IOException {
 
     URI uri = getSchema();
 
-    InputStream in = getClass().getResourceAsStream("/draft7/issue366_schema.json");
+    InputStream in = getClass().getResourceAsStream("/schema/issue366_schema.json");
```
SiemelNaran, line 38: Perplexed why this test didn't fail for me when I ran it from "mvn clean test" or right-click test folder in IntelliJ and run all tests.   Shouldn't I have seen FileNotFoundException?

gareth-robinson: getResourceAsStream seems to return null rather than throw an exception.

gareth-robinson: yeah, weird I've just noticed when running mvn that it shows
```
[INFO] Running com.networknt.schema.Issue342Test
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 s - in com.networknt.schema.Issue342Test
[INFO] Running com.networknt.schema.Issue347Test
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 s - in com.networknt.schema.Issue347Test
[INFO] Running com.networknt.schema.Issue375Test
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.001 s - in com.networknt.schema.Issue375Test
```
so it's bypassing that test??

gareth-robinson: If I rename it to Issue366FailFastTest.java, then mvn runs it.

src/test/resources/data/issue386.json
gareth-robinson: I've changed this to have all the tests above in this 'fail fast' test.
I've also updated the draft2019-09/if-then-else.json and draft7/if-then-else.json to have the full set of tests from that link too.

src/test/java/com/networknt/schema/Issue386Test.java
gareth-robinson: done!

src/test/java/com/networknt/schema/Issue366FailSlowTest.java
```diff
@@ -13,7 +13,7 @@
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
-public class Issue366FailSlow {
+public class Issue366FailSlowTest {
```
SiemelNaran, line 16: Seems these are the only two tests not ending in Test.java, so there's no more.

```
<snaran/default> [~/main/Software/json-schema-validator/src/test] [16:23:23] >> find . -name "*.java" | grep -v "Test\.java"
./java/com/networknt/schema/Issue366FailFast.java
./java/com/networknt/schema/Issue366FailSlow.java
```

SiemelNaran (commented):
LGTM, though @stevehu must approve

stevehu:
@gareth-robinson I don't know if I have merged the PR before you have updated the test cases. If that is the case, please resubmit the PR from your forked branch. The test cases are very important to ensure our code is not going to be broken by others. Thanks.

gareth-robinson:
Hi @stevehu, yeah I added the test cases in this commit: https://github.com/networknt/json-schema-validator/pull/485/commits/d7a814406aa6b7e2d42dc6b781709b5930cde5e8, seems to have been included when you merged.

stevehu:
Excellent!. Thanks a lot for your help.

PR #489, opened 2021-12-05
Fix flakiness in `CollectorContextTest`

pthariensflame:
`testCollectorContextWithKeyword` was failing under [NonDex](https://github.com/TestingResearchIllinois/NonDex) due to implementation-defined ordering of the list of context values.  This PR fixes that flakiness by sorting the list in natural order before inspecting it.


stevehu (approved):

stevehu:
@pthariensflame Thanks a lot for your help.

PR #494, opened 2022-01-12
Feature/messages.properties

huubfleuren:
all message (format-) strings are now in the resources


huubfleuren:
i see the latest version has I18nSupport

stevehu:
Yes. I think it was added by one of the contributors recently. Thanks a lot for the effort and we are all working together to make this library more useful.

Issue #446, opened 2021-09-15
maxLength and minLength do not work when a number with a zero decimal is used.


mohsin-sq:
According to the json spec 

1.0 and 1 are both considered integers, however the method used in max/minLength validations `isIntegralNumber` does not return true for 1.0. This has caused us some headaches as we are using proto3 to go from json -> proto3 and are running into the issue where when we set  maxLength value to be 2, it get converted to 2.0 and does not get evaluated by the validators: https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/MaxLengthValidator.java#L36

There was a recent issue where in jackson-databind which addressed this `isIntegralNumber` method and an `canConvertToExactIntegral` method was created for this exact purpose which should return true for both 2.0 and 2. 

We propose updating the validators (something i can also do and send a pull request) where isIntegralNumber is used in the length validators (7 locations: https://github.com/networknt/json-schema-validator/search?q=isIntegralNumber)

References:
https://json-schema.org/understanding-json-schema/reference/numeric.html
https://github.com/FasterXML/jackson-databind/issues/2885

stevehu:
@mohsin-sq  I agree with you to consider 2.0 as 2 in the maxLength and minLength validators. Would you be able to summit a PR to get it implemented? Thanks.

mohsin-sq:
I have created a PR: https://github.com/networknt/json-schema-validator/pull/450

mohsin-sq:
Thanks for merging that @stevehu! 

When do you think it'll make it into maven central?

PR #343, opened 2020-11-10
Improve type validation of numeric values

oguzhanunlu:
This PR tries to address #334 . The solution is not perfect, however, it improves the current situation.

In my opinion, we should not merge this PR as is. We should wait jackson-databind's upcoming `2.12.0` release which will [introduce](https://github.com/FasterXML/jackson-databind/blob/master/release-notes/VERSION-2.x#L26-L28) `JsonNode.canConvertToExactIntegral()`, a new method that will meet our purpose.

I wanted to open the PR so that maintainers can see my approach, the way I provide config to all validators etc.

Once jackson-databind 2.12.0 is out, I'll replace my solution with a call to `canConvertToExactIntegral`

Issue #334, opened 2020-10-14
Improve type validation of numeric values


oguzhanunlu:
**Issue**
TypeFactory uses `jackson-databind`'s `JsonNode.isIntegralNumber` and `JsonNode.isNumber` to see if the node's type is integer or number, respectively. This result is used when comparing node type and schema type which is not precise enough in some cases.

https://github.com/networknt/json-schema-validator/blob/4ba09d5f978508f7bc9e96087252333ebe90e22a/src/main/java/com/networknt/schema/TypeFactory.java#L72-L75

**Scenario**
Assume that we have a schema where a property's type is `integer`.
Given a `DecimalNode(v)` where `v` is a value that can fit into `Long`, json-schema-validator thinks that node type is `number` and this results in [type failure](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/ValidatorTypeCode.java#L63) as `number found, integer expected`

**Proposal**
Update line 72-73 in the snippet above as following (`canConvertToLong()` is defined for `DecimalNode` at `jackson-databind`)
```java
if (node.isIntegralNumber() || node.canConvertToLong())
    return JsonType.INTEGER; 
```

so that we can have a validator with better precision

stevehu:
@oguzhanunlu Thanks a lot for raising the issue and provide a solution. Let me add some background information here so that others can see where this issue is from and why we should adopt your solution as a Java-specific implementation. 

```
Warning

The precise treatment of the “integer” type may depend on the implementation of your JSON Schema validator. JavaScript (and thus also JSON) does not have distinct types for integers and floating-point values. Therefore, JSON Schema can not use type alone to distinguish between integers and non-integers. The JSON Schema specification recommends, but does not require, that validators use the mathematical value to determine whether a number is an integer, and not the type alone. Therefore, there is some disagreement between validators on this point. For example, a JavaScript-based validator may accept 1.0 as an integer, whereas the Python-based jsonschema does not.

Clever use of the multipleOf keyword (see Multiples) can be used to get around this discrepancy. For example, the following likely has the same behavior on all JSON Schema implementations:

{ "type": "number", "multipleOf": 1.0 }
```
Given the above quote from the official site, I think we should leverage the Java language feature to support the Long type; however, the change might make the validator different than the Javascript implementation. As this is a community-driven implementation, I want to know the opinion of @everybody. The best solution is to support both and use the configuration to differentiate it. We just need to make the decision of which implementation is the default.

[...]

PR #496, opened 2022-01-13
Improve type validation of integrals

christi-square:
Instead of doing string comparison, use new jackson method to determine if a number is an integral.

The `javaSemantics` config option was added in PR #343 which partially addressed issue #334. In the notes for this PR:
> Once jackson-databind 2.12.0 is out, I'll replace my solution with a call to canConvertToExactIntegral

jackson-databind has been updated to 2.12.1 so this is available but the change has not yet been made.

PR #450 which addressed #446 missed this location which is used when calling `JsonSchemaFactory.getSchema`.

Issue #344 requested coercion of various types but the only type implemented in PR #379 was lossless narrowing, set with configuration option `losslessNarrowing`. I believe that setting is unnecessary now as this implementation of `javaSemantics` addresses the original issue, but left it for backwards compatibility. It also allows you to set `javaSemantics=false` and `losslessNarrowing=true` to achieve only this specific case rather than anything that `javaSemantics` is used for in the future. At this time, these properties do exactly the same thing.

- Change from string comparison to `canConvertToExactIntegral` for `javaSemantics` and `losslessNarrowing`
- Add missing documentation around `losslessNarrowing`
- Add more test cases around integrals

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- 
+
 ## 1.0.65 - 2022-01-07
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
-- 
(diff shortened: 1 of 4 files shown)


christi-square:
@hkupty and @Navgeet may be interested in this change.

PR #500, opened 2022-01-24
Support fragment references using `$anchor`

### LINKED ISSUE #227: add anchor validator for v2019-09

Whathecode:
As is, this PR already resolves the first test case in [`V201909JsonSchemaTest.testAnchorValidator()`](https://github.com/networknt/json-schema-validator/blob/master/src/test/java/com/networknt/schema/V201909JsonSchemaTest.java#L162), but not the latter two which include absolute URIs.

It causes `java.io.FileNotFoundException: http://localhost:1234/bar` due to `"$ref": "http://localhost:1234/bar#foo"` [pointing to a local `$id`](https://github.com/networknt/json-schema-validator/blob/master/src/test/resources/draft2019-09/anchor.json#L35).

I suspect this fails for the same reason some of the other V2019_09 currently fail, e.g. [`testRefValidator()`](https://github.com/networknt/json-schema-validator/blob/master/src/test/java/com/networknt/schema/V201909JsonSchemaTest.java#L309).

It seems like this PR may already resolve the original issue I ran into (the first test case). It's less obvious how to resolve these next two errors, and whether they are related at all. Could this PR be released, as is, already? Perhaps by splitting up the test cases?

Closes #227

### DIFF
--- a/src/main/java/com/networknt/schema/JsonMetaSchema.java
+++ b/src/main/java/com/networknt/schema/JsonMetaSchema.java
@@ -368,12 +368,24 @@ public class JsonMetaSchema {
         return idNode.textValue();
     }
 
-    public String getUri() {
-        return uri;
+    public JsonNode getNodeByFragmentRef(String ref, JsonNode node) {
+        boolean nodeContainsRef = ref.equals(readId(node));
+        if (nodeContainsRef) {
+            return node;
+        } else {
+            Iterator<JsonNode> children = node.elements();
+            while (children.hasNext()) {
+                JsonNode refNode = getNodeByFragmentRef(ref, children.next());
+                if (refNode != null) {
+                    return refNode;
+                }
+            }
+        }
+        return null;
     }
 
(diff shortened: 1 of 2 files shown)


Whathecode:
I attempted to resolve the remaining failing tests, but am too unfamiliar with the codebase to make this work. Furthermore, the deeper I dig into fixing this, the more I notice this is an unrelated bug. The tests still fail even if the anchors are removed.

I will therefore consider this out of scope for this PR.

I did manage to make the second `"$ref": "http://localhost:1234/bar#foo"` test pass, by doing a recursive lookup of all containing `JsonNode`s with `$id` set and storing it in `ValidationContext`. This was then retrieved in `RefValidator.getRefSchema()`. However, this fails for the last test which has a nested non-absolute `"$id": "nested.json"`. Trying to resolve that ID was resulting in stack overflows, etc. Trying to resolve the full `id` is out of scope for how much more time I want to invest in this.

PR #501, opened 2022-01-25
oneOf_and_additionalProperties

huubfleuren:
Hi, this branch should be mergeable with the master

This issue contains a `test` a `schema` and `data` for testing the oneOf keyword in combination with addionalProperties keyword
In the test are test we expect to succeed.
Our data is a list of very different types of activities (oneOf). Some activities have their own oneOf components 

        // correct assertions would include the following 6:
        Assertions.assertTrue(errors.stream().filter(er -> er.toString().contains("toxic: is not defined in the schema")).count()==2);
        Assertions.assertTrue(errors.stream().filter(er-> er.toString().contains("$.activities[2].chemicalCharacteristic.name: is not defined in the schema")).count()==1);
        Assertions.assertTrue(errors.stream().filter(er-> er.toString().contains("$.activities[0].weight: is missing")).count()==1);

        // .. but these are unexpected fails
        Assertions.assertTrue(errors.stream().filter(er -> er.toString().contains("depth: is not defined in the schema")).count()==1);
        Assertions.assertFalse(errors.stream().filter(er-> er.toString().contains("$.activities[0].chemicalCharacteristic.categoryName: is not defined in the schema")).count()==1);
        Assertions.assertTrue(errors.stream().filter(er-> er.toString().contains("$.activities[0].heigth: number found, integer expected")).count()==1);


huubfleuren:
does not look good

PR #511, opened 2022-02-08
Add German validation messages

rustermi:
**Summary**
Similar to the addition of Chinese i18n messages, I have added German message strings.

**Drawbacks**
For simplicity reasons, I have reused the [`Issue471Test`](https://github.com/networknt/json-schema-validator/pull/472/files#diff-8ad09264f2ca6fc5a21a8995fdfe6133a8539df6dbdadadf7929faf139034598). Using a single class to test different languages would be a hassle due to the `ResourceBundle` being statically initialised. I hope this suffices as I did not want to add a Powermock dependency or do similar reflective classloader manipulation.


stevehu (approved):

AndreasALoew:
Sorry, but I seem to have run into a scenario where the translation is incomplete:

"requestBody.data darf nur für ein einziges Schema but more than one schemas {{\"$ref\":\"#/components/schemas/tEnS\"}{\"$ref\":\"#/components/schemas/vEnS\"}{\"$ref\":\"#/components/schemas/vEnSAndTens\"}} are valid  gültig sein"

This is NOT a bug in the translated messages, but rather the code location raising this validation message makes improper use of the parameter to augment the message text, which then leads to this mixture of German and English...

Can you please look into fixing this? Shall I open a new issue for this?

Thanks a million! 😄

AndreasALoew:
@stevehu Unfortunately I don't 😢 @rustermi Can you help?

My assumption is that the place in the codebase where the above exception is being thrown, adds exception text "but more than one schemas" (...) "are valid" to the exception parameters instead of treating this text as part of the message resource text that is to be translated

Does that help?

rustermi:
Good find @AndreasALoew.

I currently don't have the time to further investigate but at first glance, it seems straight forward:
[`OneOfValidator` line 287](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/OneOfValidator.java#L287) adds the untranslated message "but more than one schemas {%s} are valid ".

We need a new message string for that in
* `jsv-messages.properties`
* `jsv-messages_DE.properties`
* `jsv-messages_zh_CN.properties`

Just add a translation for messages that you speak. `ResourceBundle` will default to the English translation in `jsv-messages.properties` if there is for example no Chinese translation.

Let's say, we call this message
```yaml
oneOfMultipleSchemas = but more than one schemas {0} are valid 
```
Then, in `OneOfValidator` use `I18nSupport.getString("oneOfMultipleSchemas")`. Pay attention to the space in the end. It will potentially be removed/ignored in the messages.properties file.

Maybe this information is sufficient for you to create a PR?

AndreasALoew:
Many thanks @rustermi - from the Java codebase perspective, it now looks feasible to me.

I have to admit that this will be the very first PR that I ever raised on GitHub (we're using GitLab internally at my customer), but I hope I'll find my way through (otherwise, I'll get back here and ask...).

So stay tuned, if all ges well, you should find a PR here early next week...

AndreasALoew:
Another case of incomplete German translation:

"requestBody.data.zeitraum.vonZeitpunkt: X is an invalid date-time"

When I look into this over the weekend, I try to find the matching code location as well...

rustermi:
@AndreasALoew thanks for your efforts. You can just paste the untranslated snippet into the Github search bar on the top left and search for code appearances in this repository. You will then find that
[`ValidatorTypeCode` line 65](
https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/ValidatorTypeCode.java#L65) is untranslated as well.

AndreasALoew:
I've completed a "potential" pull request, but as I cannot get Maven tests to complete without numerous errors, I have so far refrained from creating the PR.

The issue is that I seem unable to find out how to make the Maven surefire plugin use a locale of "en_US" to run the tests (and not my default locale of de_DE)...

@rustermi can you please advise how I can successfully switch my Maven and/or JVM locale to make the Maven tests pass?

rustermi:
@AndreasALoew you can use `Locale.setDefault` to set your JVM locale. However be aware that this project initialises `ResourceBundle` statically, meaning that once it has been initialised, you can no longer switch the Locale. More specifically that means if you `setDefault` after `ResourceBundle` has been created, your changes will not have any effect.

[`Issue471Test`](https://github.com/networknt/json-schema-validator/blob/master/src/test/java/com/networknt/schema/Issue471Test.java) illustrates this perfectly. You can enable a single test and it should succeed. However, you cannot enable more than one because the Locale will only be set once.

rustermi:
The tests that verify i18n should not rely on the system locale or else they would fail on some systems and succeed on others. Instead, the i18n tests should programmatically define a locale via `Locale.setDefault` first.

I have not tried to manipulate my system locale on maven level but [this StackOverflow post](https://stackoverflow.com/a/17774592) seems to illustrate exactly this; once via pom and once via an environment variable.

You can open up a PR and assign me. If I don't get the tests running myself then, I wouldn't approve though.

AndreasALoew:
@rustermi I've tried it in several ways including the ones mentioned in the [StackOverflow post you mentioned](https://stackoverflow.com/a/17774592), but I simply cannot get the standard "mvn clean install" surefire tests to assume any other locale than German, and so a huge number fails...

Either there is something really strange in my environment that does not allow Locales to be overridden (but I would have no idea what this could be...), or I fear that the whole "json-schema-validator" testing needs to be reworked to support running from clients with a default locale that is different from en_US... 😢 

So if you also don't know of a specific change in Maven settings and/or MAVEN_OPTS, JVM system properties etc. I fear that the only way to move forward is that I create an untested PR (or attach an old-style unified patch, because it's just very few lines...).

Please advise!

rustermi:
Do so, I'm working with an English locale anyways. If tests run green for me, your changes look okay and I understand why it would fail on your machine, I'd approve it.

To help, it would be great if you could post your (failing test) output together with your PR so that I can verify what's going on.

AndreasALoew:
ALL CLEAR: Successfully uncovered and properly resolved the true root cause of the Locale/testing issues -> see PR #536

rishabh413:
Hi, I am importing this library for my project and I am trying to understand why System.exit() calls were added in this commit. In general, a library is not expected to have JVM exit() calls. Could you help understand the requirements here?

stevehu:
I did a full-text search and couldn't find the exit() you mentioned. Could you please point me to the line of code? Thanks.

rishabh413:
Thanks for looking into it.

1) PTAL at search results here:
screenshot

2) The code is in `I18nSupport` class: [1](https://github.com/networknt/json-schema-validator/blob/0b51d257a974c84c55f00f15a768bb463dca2e75/src/main/java/com/networknt/schema/I18nSupport.java#L25) [2](https://github.com/networknt/json-schema-validator/blob/0b51d257a974c84c55f00f15a768bb463dca2e75/src/main/java/com/networknt/schema/I18nSupport.java#L39)

3) These were added in `Add German validation messages #511` screenshot

stevehu:
@rishabh413 Thanks for raising it. I have fixed it with the following issue. 

https://github.com/networknt/json-schema-validator/issues/708

Issue #526, opened 2022-02-28
Support for non-existent $id URLs?


davidpricedev:
My json-schema has `$id`s that aren't "real" urls that can be requested. I also use json-schema bundling.

It seems this library doesn't support bundled schemas - at least it doesn't follow the `$ref` to the `$defs` - it still tries to make requests out to the web to fetch the `$ids` of the items that are in the `$defs` section.

I've been able to work around this sort of limitation in other libraries by one of these tactics: 

- pre-loading all the individual json-schema documents into an internal storage provided by the library
- changing out the document fetching with a custom component that can do a file-system lookup

Does this library support either of those options? I've had a look through the docs and looked around a bit in the code too, but nothing is jumping out at me.

stevehu:
@davidpricedev I am wondering if you could submit a PR with a test case to reproduce the issue? I need to see what is your schema and what is your json. We are using the bundler all the time and there is no issue at all. We have an [openapi-bundler](https://github.com/networknt/openapi-bundler) in the networknt organization.

davidpricedev:
@stevehu thanks for taking a look. I have a PR out now with a reproduction of the issue.

claussni:
Any progress on this? I'm stuck with the issue. Can't seem to hack a workaround.

PR #527, opened 2022-02-28
Issue 526 reproduction

davidpricedev:
This reproduces the issue described in #526.

I'd expect the example json document to validate against the bundled schema, but it fails with http errors trying to load example.com instead. 

> 15:06:40.995 [main] ERROR c.networknt.schema.JsonSchemaFactory - Failed to load json schema!
java.net.UnknownHostException: schemas.example.com
...

### DIFF
--- a/src/test/resources/schema/issue526.schema.json
+++ b/src/test/resources/schema/issue526.schema.json
@@ -1,0 +1,64 @@
+{
+  "$id": "https://schemas.example.com/example-service/commands/1.0.0/example.schema.json",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "title": "example",
+  "type": "object",
+  "properties": {
+    "phones": {
+      "description": "The example phone numbers",
+      "type": "array",
+      "items": {
+        "$id": "https://schemas.example.com/example-service/objects/1.0.0/phone.schema.json",
+        "description": "try an absolute $ref path"
+      }
+    },
+    "email_addresses": {
+      "description": "The example email address",
+      "type": "array",
+      "items": {
+        "$ref": "../../objects/1.0.0/email-address.schema.json",
+        "description": "try a relative $ref path"
+      }
+    }
+  },
+  "required": [
+    "email_addresses",
+    "phones"
+  ],
+  "additionalProperties": false,
+  "$defs": {
+    "https://schemas.example.com/example-service/objects/1.0.0/phone.schema.json": {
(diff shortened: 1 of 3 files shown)


src/test/resources/schema/issue526.schema.json
```diff
@@ -0,0 +1,64 @@
...
+  "$id": "https://schemas.example.com/example-service/commands/1.0.0/example.schema.json",
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "title": "example",
+  "type": "object",
+  "properties": {
+    "phones": {
+      "description": "The example phone numbers",
+      "type": "array",
+      "items": {
+        "$ref": "https://schemas.example.com/example-service/objects/1.0.0/phone.schema.json",
```
josejulio, line 11: For refs that are inside the same file, we use it like this:
`"$ref": "#/definitions/Policy"`

with the definition looking like:

```
"definitions": {
    "Policy": {
      "type": "object",
      "properties": {
        "id": {
      ...
```

The `#` is used for stuff relative to the same file.

The full file is [here](https://github.com/RedHatInsights/event-schemas/blob/main/schemas/apps/policies/v1/policy-triggered.json#L12).

If you want to use ids that are not real urls, you could implement an URI factory like:

```java
static class CloudEventURIFactory implements URIFactory {

        private static final String baseUrl = "https://console.redhat.com/api";
        private final URLFactory urlFactory = new URLFactory();
        private final String base;

        CloudEventURIFactory() {
            String fullPath = RHELSystem.class.getResource(schemaPath).toString();
            base = fullPath.substring(0, fullPath.length() - schemaPath.length());
        }

        @Override
        public URI create(String uri) {
            return urlFactory.create(replaceBase(uri));
        }

        @Override
        public URI create(URI baseURI, String segment) {
            return urlFactory.create(URI.create(replaceBase(baseURI.toString())), segment);
        }

        private String replaceBase(String uri) {
            if (uri.startsWith(baseUrl)) {
                uri = base + uri.substring(baseUrl.length(), uri.length() - 1);
            }

            return uri;
        }
    }
```

and adding it to your JsonSchemaFactory: 

```
JsonSchemaFactory.Builder().
                .uriFactory(new CloudEventURIFactory(), "https")
                .build();
```

Full example here: https://github.com/RedHatInsights/event-schemas-java/blob/main/src/main/java/com/redhat/cloud/event/parser/ConsoleCloudEventParser.java#L35C15-L63

I hope that above information is useful, I didn't go a lot into the details.

stevehu:
Issue #526 has been resolved. Thanks.

PR #529, opened 2022-03-09
Updates to German translation

rustermi:
Since we now have the newest version (1.0.67) running in production, I found a few things this PR fixes now:

1. **Renamed `messages_de_DE.properties`**

    The translation is understandable for all German speaking people. However, the previous file name lead to these translations only being used for Germany (`de_DE` — German on German territory). A JVM running with Austrian German (`de_AT`) or simply German without any territory descriptor (`de`) resulted in the English default messages being used.

1. **Update German translations**

    - 'false' translation was missing
    - Wording has been improved for cases with zero, one or many elements
    - Capitalisation of words after a colon has been unified

Sorry for the inconvenience and thank you for your work.


stevehu:
@rustermi Thanks a lot for your help.

PR #533, opened 2022-03-14
fixes #532 Invalid (non-string) $schema produces NullPointerException

### LINKED ISSUE #532: Invalid (non-string) $schema produces NullPointerException
The following code produces a `NullPointerException`:

```java
JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
factory.getSchema("{ \"$schema\": true }");
```

The `$schema` property should be a string but there is no check that it is, in fact, a string.

Note that I have a fix for this I will put up in a PR shortly.

christi-square:


stevehu (approved):

stevehu:
@christi-square Thanks a lot for your help.

PR #534, opened 2022-03-17
Adding Unevaluated properties keyword.

prashanthjos:
This PR adds unevaluatedProperties keyword implementation to networknt's json-schema-validator. Please refer to the following discussion for more details on how this keyword might work.

https://github.com/json-schema-org/json-schema-spec/issues/556


prashanthjos:
@stevehu can you please have a look at this PR and give your suggestions.

stevehu:
@prashanthjos Thanks a lot for the PR. It was a good read for the spec :) I understand that the keyword is introduced in 2020-12 and we don't have this version created yet. I am going to find some time to add this new version. Meanwhile, I am wondering if you could add a test case to use the standard test suite? The test case can help us to keep the code stable if someone else wants to contribute to this keyword. Here is the link to the test https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/master/tests/draft2020-12/unevaluatedProperties.json

prashanthjos:
Sure @stevehu I will try to run the tests mentioned in the URL. Also please note that this unevaluatedProperties PR is only applicable for a boolean value for this keyword and not a schema.

stevehu:
@TJC Thanks a lot for identifying the breaking point. Let's review the changes in this PR again to see if we can find out the root cause. @prashanthjos I am wondering if you could help to review the code. If necessary, I can set up a zoom meeting to work together.

TJC:
Andreas helped me make a heap dump. Looking at the profiler now, it kinda looks like the `AnyOfValidator.java` has an ArrayList on line 80 which is not being cleared at the end of a validation, and just keeps accumulating more and more elements as more files get validated.

PR #539, opened 2022-03-25
Refactoring-code

Sahil3198:
Here I made some changes to the code to improve code readability. 
I used some refactoring Techniques such as Extract Method, Rename variable, Pull-up variable, Extract class, Replace conditional with polymorphism, and Push-down method.

### DIFF
--- a/src/main/java/com/networknt/schema/AllOfValidator.java
+++ b/src/main/java/com/networknt/schema/AllOfValidator.java
@@ -31,7 +31,6 @@ import org.slf4j.LoggerFactory;
 public class AllOfValidator extends BaseJsonValidator implements JsonValidator {
     private static final Logger logger = LoggerFactory.getLogger(AllOfValidator.class);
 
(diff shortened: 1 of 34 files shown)


stevehu (approved):

stevehu:
@Sahil3198 Thanks a lot for your help. I didn't realize that we have the validationContext in almost all the validators as it was introduced for only one or two validators in the beginning and almost every validator is using it these days. It makes perfect sense to move to the BaseJsonValidator. Also, extracting the different versions of the implementation paves the road for use to support new versions easily.

TJC:
Release 1.0.68 introduced a nasty memory leak. There were only a couple of significant PRs included in that release, and this was one of them. See #546 for details. I'm not really sure how this PR could have triggered it, but thought I'd mention the problem.

TJC:
Ah, actually my git bisect points to this commit 500724272bb887bc25c951b1d8cedfae742edec7 as the cause.

Issue #538, opened 2022-03-25
Cannot get properties validator from map due to custom comperator


0x4a616e:
If you call `JsonSchema.getValidators()` you'll get a `Map<String, JsonValidator>`. However, you cannot get the properties validator out of it, as the map uses a custom comparator. So `JsonSchema.getValidators().get("#/properties")` always returns `null` even though the key exists. The culprit seems to be this:

```java
    private static Comparator<String> VALIDATOR_SORT = (lhs, rhs) -> {
        if (lhs.endsWith("/properties")) {
            return -1;
        }
        if (rhs.endsWith("/properties")) {
            return 1;
        }
        return lhs.compareTo(rhs);
    };
```

I guess a solution could be to return `0` if `lhs.equals(rhs)` before checking for the "/properties" suffix.

stevehu:
@0x4a616e It sounds reasonable. Could you please submit a PR with your test cases? We just want to make sure the fix is backward compatible. Thanks.

0x4a616e:
@stevehu I created a PR with the change mentioned above: https://github.com/networknt/json-schema-validator/pull/541

PR #541, opened 2022-03-29
Allow fetching /properties from map with comparator

0x4a616e:
Fixes https://github.com/networknt/json-schema-validator/issues/538


stevehu:
@0x4a616e Thanks a lot for your help.

Issue #550, opened 2022-04-07
[Question] Does this tool provide Schema path?


ymszzq:
I see ValidationMessage provide the path to the input json. I couldn't find the path to schema at where the error throws.

Can you please confirm if I understand it correctly?

Are there any plans to support this? I think it is quite important.

ymszzq:
I can submit a PR

stevehu:
@ymszzq Thanks a lot for your help. The PR is merged and will be in the next release.

ymszzq:
My pleasure!

When will you release next time?

Thanks

stevehu:
The PR was merged in April, and we have released several versions since then.

PR #552, opened 2022-04-08
Add schemaPath to ValidationMessage.

ymszzq:
https://github.com/networknt/json-schema-validator/issues/550

### DIFF
--- a/src/main/java/com/networknt/schema/ValidationMessage.java
+++ b/src/main/java/com/networknt/schema/ValidationMessage.java
@@ -26,6 +26,7 @@ public class ValidationMessage {
     private String type;
     private String code;
     private String path;
+    private String schemaPath;
     private String[] arguments;
     private Map<String, Object> details;
     private String message;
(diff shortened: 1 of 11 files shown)


src/main/java/com/networknt/schema/ValidationMessage.java
```diff
@@ -113,17 +130,17 @@ public void setType(String type) {
         this.type = type;
     }
 
-    public static ValidationMessage of(String type, ErrorMessageType errorMessageType, String at, String... arguments) {
+    public static ValidationMessage of(String type, ErrorMessageType errorMessageType, String at, String schemaPath, String... arguments) {
```
rphilipp64, line 133: This is a breaking change, this should at least have been a bump of the minor version!

What's even worse, with the rest `arguments` parameter also being of the same type (`String`), any call to `ValidationMessage.of` that includes at least one argument **silently** shifts the first rest argument to become the `schemaPath` instead. No compiler warnings or anything to help detect this. Since this requires a manual search, I feel the [changelog entry](https://github.com/networknt/json-schema-validator/blob/4906d90c6f3c2dc63646bb09f0a0bdc11b6cd073/CHANGELOG.md?plain=1#L24) could have been a bit more detailed about this too.

I had some fun finding this out going from `1.0.68` → `1.0.69`, I was lucky to have a test fail on it to detect the arguments shift…

Issue #183, opened 2019-09-04
Validation error when field is nullable and consumer sends in a null value


ddobrin:
When an OpenAPI specification declares an enum, and specifies it as being nullable, the consumer can send the values:
{...
field: null
...}

The validator identifies it correctly as nullable, however proceeds to look for null as a value in the enumerated fields.

It is a delta between OpenAPI and JSONSchema specifications:
-- https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values
-- http://spec.openapis.org/oas/v3.0.2

The issue can be fixed in a configurable manner in the validator without impacting performance

stevehu:
We have a configuration for this type of use cases already. 

https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/SchemaValidatorsConfig.java

harishvashistha:
@stevehu I could not see above configuration working for this example. We have recently onboarded with this library for schema validation with some existing migrated data and having issues with earlier validated resources. Can you help me with these example:

```
"schema": {
      "properties": {
        "values": {
          "type": "string",
          "nullable": true,
          "default": null
        }
      }
    }
```
Data:

```
{
    "values": null
 }
```

It is failing with this exception:

`null found, string expected`

Above handleNullableFields seems working only with oneOf and enum. Urgent help needed.

stevehu:
You have to implicitly set the typeLoose flag to true when you initialize.

harishvashistha:
@stevehu I have set this configuration for `SchemaValidatorsConfig` :

```
@Bean
    public SchemaValidatorsConfig schemaValidatorsConfig() {
        ApplyDefaultsStrategy applyDefaultsStrategy =
                new ApplyDefaultsStrategy(true, true, true);
        SchemaValidatorsConfig schemaValidatorsConfig = new SchemaValidatorsConfig();
        schemaValidatorsConfig.setTypeLoose(true);
        schemaValidatorsConfig.setApplyDefaultsStrategy(applyDefaultsStrategy);
        return schemaValidatorsConfig;
    }
```
[...]

Issue #553, opened 2022-04-13
default as null is not being set


harishvashistha:
Hi Team,

if schema definition specifies that `"default":null ` when that property is not specified in json resource then why there is this check in code which sets default values only when those values are non-null. You are requested to please think about below section of code when we want default = null should be set when no value is provided explicitly. e.g.

```
"schema": {
      "properties": {
        "values": {
          "type": "string",
          "nullable": true,
          "default": null
        },
       "id":{
           "type": "string",
          "nullable": true,
          "default": "US-12"
        }
      }
    }
```
Resource to be created:

```
"data":{

}
```
we want `public ValidationResult walk(JsonNode node, boolean shouldValidateSchema)` in `JsonSchema.class` should result into this:

```
"data":{
   "values": null,
   "id":"US-12"
}
```
But it actually returns this:
```
"data":{
   "id":"US-12"
}
```

which is not acceptable when this payload is provided to those languages where having null value has different meaning than having non existence of key altogether. You are requested to please help here. As this code is affecting us:

```
private void applyPropertyDefaults(ObjectNode node) {
        for (Map.Entry<String, JsonSchema> entry : schemas.entrySet()) {
            JsonNode propertyNode = node.get(entry.getKey());

            if (propertyNode == null || (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull())) {
                JsonSchema propertySchema = entry.getValue();
                JsonNode defaultNode = propertySchema.getSchemaNode().get("default");
                if (defaultNode != null && !defaultNode.isNull()) {
                    // mutate the input json
                    node.set(entry.getKey(), defaultNode);
                }
            }
        }
    }
```

in propertiesValidator.java file

stevehu:
Please take a look at this class. 

https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/ApplyDefaultsStrategy.java

harishvashistha:
@stevehu I have set this configuration for `SchemaValidatorsConfig` :

```
@Bean
    public SchemaValidatorsConfig schemaValidatorsConfig() {
        ApplyDefaultsStrategy applyDefaultsStrategy =
                new ApplyDefaultsStrategy(true, true, true);
        SchemaValidatorsConfig schemaValidatorsConfig = new SchemaValidatorsConfig();
        schemaValidatorsConfig.setTypeLoose(false);
[...]

PR #554, opened 2022-04-14
removed unnecessary check which causes nullable=true, handleNullableFields= true with value=null to fail in all cases except OneOf

harishvashistha:
when nullable = true , handlenullableFields = true and value supplied = null , then getting this failure message `"null found, string expected"` for this example:

```
 "schema": {
      "properties": {
        "values": {
          "type": "string",
          "nullable": true,
          "default": null
        }
      }
    }
```

Data to be created:

```
"data": {
          "values": null
        }
```

Reason appears this unnecessary check for `state.isComplexValidator()` which is true only in case of oneOf . so it suppresses all handleNullableFields configurations causing the above error message.

This is related to my latest query on this issue https://github.com/networknt/json-schema-validator/issues/183

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.networknt</groupId>
     <artifactId>json-schema-validator</artifactId>
-    <version>1.0.68</version>
+    <version>1.0.68-SNAPSHOT1</version>
     <packaging>bundle</packaging>
     <description>A json schema validator that supports draft v4, v6, v7 and v2019-09</description>
     <url>https://github.com/networknt/json-schema-validator</url>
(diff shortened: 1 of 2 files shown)


stevehu:
@harishvashistha Thanks for your PR. I have run the test cases and they all passed from your branch. I cannot remember why we checked the complex validator in the first place but I guess there was a reason for that. I am wondering if we should handle the NULL for all validators or only a limited set of validators.

harishvashistha:
@stevehu @prashanthjos should I merge this PR into another PR that is open https://github.com/networknt/json-schema-validator/pull/555 , which I created for another issue https://github.com/networknt/json-schema-validator/issues/553  ?

We need code changes of both the PRs in your master library at the same time so that we can plan our product release bug free.

stevehu:
No worry. We are looking at both PRs together. Thanks.

harishvashistha:
@SiemelNaran @stevehu @prashanthjos can we get these changes in master release please. This PR is urgently required with PR https://github.com/networknt/json-schema-validator/pull/555 as we are planning a relase on Tuesday.
I hope you would understand the urgency for these changes.

harishvashistha:
@prashanthjos You are requested to please approve and merge this PR and if there are some review comments I can try to add and fix those quickly. This is the highest priority thing I am stuck with in my project.
I hope you would understand.

PR #555, opened 2022-04-14
Setting default value even if that value is null

harishvashistha:
Hi Team,

We have a use case as mention in this issue https://github.com/networknt/json-schema-validator/issues/553 which is causing failures in downstream systems where "x":null has different meaning than non existance of "x". can you please evaluate if this change can go in future release else we would have to put some hacky code before calling json-schema-validator which modifies the json which is going to be validated and thereby breaks the purpose of this library.

Please help

### DIFF
--- a/src/main/java/com/networknt/schema/ItemsValidator.java
+++ b/src/main/java/com/networknt/schema/ItemsValidator.java
@@ -123,9 +123,6 @@ public class ItemsValidator extends BaseJsonValidator implements JsonValidator {
             JsonNode defaultNode = null;
             if (applyDefaultsStrategy.shouldApplyArrayDefaults() && schema != null) {
                 defaultNode = schema.getSchemaNode().get("default");
-                if (defaultNode != null && defaultNode.isNull()) {
-                    defaultNode = null;
-                }
             }
             int i = 0;
             for (JsonNode n : arrayNode) {
--- a/src/main/java/com/networknt/schema/PropertiesValidator.java
+++ b/src/main/java/com/networknt/schema/PropertiesValidator.java
@@ -137,7 +137,7 @@ public class PropertiesValidator extends BaseJsonValidator implements JsonValida
                 continue;
             }
             boolean applyDefault = propertyNode == null
-                    || (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull() && !defaultNode.isNull());
+                    || (propertyNode.isNull() && applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull());
             if (applyDefault) {
                 node.set(entry.getKey(), defaultNode);
             }
--- a/src/main/java/com/networknt/schema/PropertiesValidator.java
+++ b/src/main/java/com/networknt/schema/PropertiesValidator.java
@@ -135,7 +135,7 @@ public class PropertiesValidator extends BaseJsonValidator implements JsonValida
             if (propertyNode == null || (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull())) {
                 JsonSchema propertySchema = entry.getValue();
                 JsonNode defaultNode = propertySchema.getSchemaNode().get("default");
-                if (defaultNode != null && !defaultNode.isNull()) {
+                if (defaultNode != null) {
                     // mutate the input json
                     node.set(entry.getKey(), defaultNode);
                 }
--- a/src/main/java/com/networknt/schema/PropertiesValidator.java
+++ b/src/main/java/com/networknt/schema/PropertiesValidator.java
@@ -132,15 +132,25 @@ public class PropertiesValidator extends BaseJsonValidator implements JsonValida
         for (Map.Entry<String, JsonSchema> entry : schemas.entrySet()) {
             JsonNode propertyNode = node.get(entry.getKey());
 
-            if (propertyNode == null || (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull())) {
-                JsonSchema propertySchema = entry.getValue();
-                JsonNode defaultNode = propertySchema.getSchemaNode().get("default");
+            if (propertyNode == null){
+                JsonNode defaultNode = getDefaultNode(entry);
                 if (defaultNode != null) {
+                    // mutate the input json
+                    node.set(entry.getKey(), defaultNode);
+                }
+            }else if (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull()){
+                JsonNode defaultNode = getDefaultNode(entry);
+                if (defaultNode != null && !defaultNode.isNull()) {
                     // mutate the input json
                     node.set(entry.getKey(), defaultNode);
                 }
             }
         }
+    }
+
+    private JsonNode getDefaultNode(final Map.Entry<String, JsonSchema> entry) {
+        JsonSchema propertySchema = entry.getValue();
+        return propertySchema.getSchemaNode().get("default");
     }
 
     private void walkSchema(Map.Entry<String, JsonSchema> entry, JsonNode node, JsonNode rootNode, String at,
--- a/src/main/java/com/networknt/schema/PropertiesValidator.java
+++ b/src/main/java/com/networknt/schema/PropertiesValidator.java
@@ -132,18 +132,14 @@ public class PropertiesValidator extends BaseJsonValidator implements JsonValida
         for (Map.Entry<String, JsonSchema> entry : schemas.entrySet()) {
             JsonNode propertyNode = node.get(entry.getKey());
 
-            if (propertyNode == null){
-                JsonNode defaultNode = getDefaultNode(entry);
-                if (defaultNode != null) {
-                    // mutate the input json
-                    node.set(entry.getKey(), defaultNode);
-                }
-            }else if (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull()){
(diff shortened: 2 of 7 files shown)


stevehu:
@harishvashistha I think this PR covers the previous one. Could we close that one and only leave this one open? Thanks.

harishvashistha:
PR https://github.com/networknt/json-schema-validator/pull/554/files has changes related to file

`src/main/java/com/networknt/schema/utils/JsonNodeUtil.java`

so that nullable fields succeeds when handleNullable is true and null value supplied. This current PR does not have that change.
This PR is related to setting null as default value when `"default": "null"` is specified in schema definition.

Do you want me to merge these two PRs into this one for tracking?

stevehu:
No necessary. You are using the same branch to track both PRs so they merged already automatically. If you look at this PR, there are three commits.

harishvashistha:
@stevehu apologies for confusion, but this PR does not have code from another PR as i reverted that code manually while creating this branch, as I wanted both the PRs to show relevant changes only for the issue they are targeting and mentioned in subject. Please consider both the PRs with respect to the file changes they have. 
When can we expect these two PRs in official master as we are planning our project release to production soon?

stevehu:
The changes look good to me. I just want other developers to have some input on the changes as they are more familiar with this feature. Once merged, we are expecting a release soon.

src/main/java/com/networknt/schema/PropertiesValidator.java
```diff
@@ -135,7 +135,7 @@ private void applyPropertyDefaults(ObjectNode node) {
             if (propertyNode == null || (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull())) {
                 JsonSchema propertySchema = entry.getValue();
                 JsonNode defaultNode = propertySchema.getSchemaNode().get("default");
-                if (defaultNode != null && !defaultNode.isNull()) {
+                if (defaultNode != null) {
                     // mutate the input json
                     node.set(entry.getKey(), defaultNode);
                 }
```
SiemelNaran, line 141: In ItemsValidator they have similar code
```
                if (defaultNode != null && defaultNode.isNull()) {
                    defaultNode = null;
                }
```
Maybe this should be removed?

SiemelNaran: Can you add unit test to JsonWalkApplyDefaultsTest?  You'll have to change walk-schema-default.json, adding `"default": null` somewhere, and similar changes to walk-data-default.json.

SiemelNaran (changes requested):

src/main/java/com/networknt/schema/PropertiesValidator.java
harishvashistha: > In ItemsValidator they have similar code
> 
> ```
>                 if (defaultNode != null && defaultNode.isNull()) {
>                     defaultNode = null;
>                 }
> ```
> 
> Maybe this should be removed?

In ItemsValidator, this code is fine as Library specification says `applyArrayDefaults if true then apply defaults inside json arrays if the attribute is explicitly null` so code in here is checking if incoming item is null and default is not null then set that not null item in the array in place of null as there is no point of setting null in place of null. This is depicted in this code:

```
for (JsonNode n : arrayNode) {
                if (n.isNull() && defaultNode != null) {
                    arrayNode.set(i, defaultNode);
                    n = defaultNode;
                }
                doWalk(validationMessages, i, n, rootNode, at, shouldValidateSchema);
                i++;
            }
```

harishvashistha: > Can you add unit test to JsonWalkApplyDefaultsTest? You'll have to change walk-schema-default.json, adding `"default": null` somewhere, and similar changes to walk-data-default.json.

I have added some required test cases, Please have a look.

src/test/resources/schema/walk-schema-default.json
```diff
@@ -10,6 +10,20 @@
...
+        },
+        "stringValue_provided_as_null_with_default_null": {
+          "description": "the test data has this attribute as null, so default as null should be applied if the ApplyDefaultsStrategy requires it",
+          "type": "string",
+          "default": null
+        },
+        "stringValue_missing_with_no_default": {
+          "description": "the test data does not have this attribute, no default value should be applied",
+          "type": "string"
         }
```
SiemelNaran, line 27: Do you need lines 19-27? It seems lines 14-18 is good enough.

src/main/java/com/networknt/schema/PropertiesValidator.java
SiemelNaran: > In ItemsValidator, this code is fine as Library specification says `applyArrayDefaults if true then apply defaults inside json arrays if the attribute is explicitly null` so code in here is checking if incoming item is null and default is not null then set that not null item in the array in place of null as there is no point of setting null in place of null. This is depicted in this code:

This makes sense.  However at some point I was thinking we may want to add an event listener, that's why I created the ApplyDefaultsStrategy class to add more stuff in the future if we want to.  Now we may want the listener to count the number of places where the default is missing or something else, so replacing null with null and firing the listener makes sense in that case.  What do you think?

src/main/java/com/networknt/schema/PropertiesValidator.java
```diff
@@ -132,9 +132,14 @@ private void applyPropertyDefaults(ObjectNode node) {
...
+            if (propertyNode == null){
+                JsonNode defaultNode = getDefaultNode(entry);
+                if (defaultNode != null) {
+                    // mutate the input json
+                    node.set(entry.getKey(), defaultNode);
+                }
+            }else if (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull()){
+                JsonNode defaultNode = getDefaultNode(entry);
                 if (defaultNode != null && !defaultNode.isNull()) {
                     // mutate the input json
```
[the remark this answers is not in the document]
harishvashistha: make sense, updating accordingly

src/test/resources/data/walk-data-default.json
```diff
@@ -7,6 +7,7 @@
     "goodArray": ["hello", null],
     "badArray": ["hello", null],
     "reference": {
+      "stringValue_provided_as_null_with_default_null": null
```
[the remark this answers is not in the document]
harishvashistha: I applied this test case just to cover some if/else condition I have put in place. But I have removed it to be according to my description of changes

src/test/resources/schema/walk-schema-default.json
harishvashistha: @SiemelNaran I removed one unnecessary schema case. But other schema case I have left from 19-21 is covering code ranging from 136 to 138 in PropertiesValidator.java file. I think this case makes sense.

harishvashistha:
@SiemelNaran I have fixed all the review comments. Please try to release this code with PR https://github.com/networknt/json-schema-validator/pull/554 as we urgently need these fixes for our Tuesday release next week in production.

src/main/java/com/networknt/schema/ItemsValidator.java
```diff
@@ -123,9 +123,6 @@ public Set<ValidationMessage> walk(JsonNode node, JsonNode rootNode, String at,
             JsonNode defaultNode = null;
             if (applyDefaultsStrategy.shouldApplyArrayDefaults() && schema != null) {
                 defaultNode = schema.getSchemaNode().get("default");
-                if (defaultNode != null && defaultNode.isNull()) {
-                    defaultNode = null;
-                }
```
SiemelNaran, line 128: This part was an open question.  If you are removing these lines, then it means you are applying default even if it is null in the case of a json array.  So then for parallel it make sense to do this for the other file where the
```
            boolean applyDefault = propertyNode == null
                    || (applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull() && propertyNode.isNull() && !defaultNode.isNull());
```
should be changed to
```
            boolean applyDefault = propertyNode == null || propertyNode.isNull()  && applyDefaultsStrategy.shouldApplyPropertyDefaultsIfNull();
```

On the other hand if you want to skip applying default if the default is null node, then keep line lines the other fille as is but bring back lines 126-128. 

Will let you decide how to want to go, but keep things in parallel.

SiemelNaran (approved):
Approved after minor comment addressed.  I'll review again afterwards.

src/main/java/com/networknt/schema/ItemsValidator.java
harishvashistha: @SiemelNaran I have aligned the logic at both the places, by setting null even if value provided is null, because of your listener logic that may come in future for some requirement.

harishvashistha:
@SiemelNaran @stevehu @prashanthjos can we get these changes in master release please. This PR is urgently required with PR https://github.com/networknt/json-schema-validator/pull/554 as we are planning a relase on Tuesday.
I hope you would understand the urgency for these changes.

stevehu:
@harishvashistha I am not familiar with the feature and depend on other team members to review the changes. Once the PR is reviewed and merged, we will have a release.

harishvashistha:
@stevehu thanks for the update, may be if I can ask @SiemelNaran  and @prashanthjos if you guys can help me with these two PRs, it would be a great gesture on your side.
I am fully dependent on this library to move to production with our project. Please help with these PRs: https://github.com/networknt/json-schema-validator/pull/554 , https://github.com/networknt/json-schema-validator/pull/555

SiemelNaran (approved):

SiemelNaran:
This PR looks good to me.  I don't know if I have the ability to merge to master.  I will try now.

stevehu (approved):

Issue #535, opened 2022-03-18
OneOf validation gives unnecessary errors


RenegadeWizard:
Let's say we have a schema like this one:
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "https://example.com/product.schema.json",
  "title": "",
  "description": "",
  "type": "object",
  "properties": {
    "mainprop": {
      "type": "object",
      "oneOf": [
        {
          "required": [
            "property1"
          ],
          "properties": {
            "property1": {
              "enum": [
                "AAA",
                "BBB",
                "CCC"
              ],
              "type": "string"
            }
          }
        },
        {
          "required": [
            "property2"
          ],
          "properties": {
            "property2": {
              "type": "number"
            }
          }
        }
      ]
    }
  },
  "required": [ "mainprop" ]
}
```
So we have one `oneOf` property that requires either `property1` or `property2`. 
```json
{
  "mainprop": {
    "property1": "INVALID"
  }
}
```
This json is incorrect and the validator should return error that the value of `property1` is incorrect, but in addition to that it returns an error that property2 is required and missing:
```
$.mainprop.property1: does not have a value in the enumeration [AAA, BBB, CCC]
$.mainprop.property2: is missing but it is required
```
It's not correct, because `property1` is already present, just incorrect. So the error that property2 is required is unnecessary/incorrect.

I have a fix for that, just have to submit a PR

AndreasALoew:
I think I have seen the same issue recently.

Would it be possible that we do see the superfluous "is missing but it is required" not just in an OneOf, but also an AnyOf scenario (with at least one value required, but more than one valid)?

Looking forward to seeing your PR submitted...

RenegadeWizard:
Sure, I would be happy to submit the PR, but I don't have the permissions to push my changes to a branch. Can you help with that?

AndreasALoew:
@RenegadeWizard , you can perfectly create a PR from your own fork (and typically also, a dedicated branch in this fork) of the project: see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork

RenegadeWizard:
Great, thanks for your help, I submitted a PR: #537

AndreasALoew:
Hello @RenegadeWizard,

[...]

PR #559, opened 2022-04-24
proposal for issue #535, part 2: fix the same issue in AnyOfValidator, too

AndreasALoew:
As stated in https://github.com/networknt/json-schema-validator/issues/535, the same issue with receiving superfluous validation messages of type "required" after at least one subschema has validated without error does also apply to AnyOfValidator.

Here's my proposed fix for this second part... 😄 

@RenegadeWizard and @stevehu : please validate - many thanks! 👍

### DIFF
--- a/src/main/java/com/networknt/schema/AnyOfValidator.java
+++ b/src/main/java/com/networknt/schema/AnyOfValidator.java
@@ -70,6 +70,7 @@ public class AnyOfValidator extends BaseJsonValidator implements JsonValidator {
         CollectorContext.getInstance().add(UnEvaluatedPropertiesValidator.EVALUATED_PROPERTIES, new ArrayList<>());
 
         try {
+            int numberOfValidSubSchemas = 0;
             for (JsonSchema schema : schemas) {
                 Set<ValidationMessage> errors = new HashSet<>();
                 if (schema.getValidators().containsKey(typeValidatorName)) {
@@ -86,6 +87,17 @@ public class AnyOfValidator extends BaseJsonValidator implements JsonValidator {
                 } else {
                     errors = schema.walk(node, rootNode, at, true);
                 }
+                
+                // check if any validation errors have occurred
+                if (errors.isEmpty()) {
+                    // check whether there are no errors HOWEVER we have validated the exact validator
+                    if (!state.hasMatchedNode()) {
+                        continue;
+                    }
+                    // we found a valid subschema, so increase counter
+                    numberOfValidSubSchemas++;
+                }                
+                
                 if (errors.isEmpty() && (!this.validationContext.getConfig().isOpenAPI3StyleDiscriminators())) {
                     // Clear all errors.
                     allErrors.clear();
@@ -104,6 +116,14 @@ public class AnyOfValidator extends BaseJsonValidator implements JsonValidator {
                     }
                 }
                 allErrors.addAll(errors);
+            }
+
+            // determine only those errors which are NOT of type "required" property missing
+            Set<ValidationMessage> childNotRequiredErrors = allErrors.stream().filter(error -> !ValidatorTypeCode.REQUIRED.getValue().equals(error.getType())).collect(Collectors.toSet());
+
+            // in case we had at least one (anyOf, i.e. any number >= 1 of) valid subschemas, we can remove all other errors about "required" properties
+            if (numberOfValidSubSchemas >= 1 && childNotRequiredErrors.isEmpty()) {
+                allErrors = childNotRequiredErrors;
             }
 
             if (this.validationContext.getConfig().isOpenAPI3StyleDiscriminators() && discriminatorContext.isActive()) {
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.networknt</groupId>
     <artifactId>json-schema-validator</artifactId>
-    <version>1.0.69</version>
+    <version>1.0.70-SNAPSHOT</version>
     <packaging>bundle</packaging>
     <description>A json schema validator that supports draft v4, v6, v7 and v2019-09</description>
     <url>https://github.com/networknt/json-schema-validator</url>


AndreasALoew:
Please note that the **current snapshot of master has some test failures** - but they are NOT due to this my PR:

```
[ERROR] Errors:
[ERROR]   V201909JsonSchemaTest.testRefRemoteValidator:315->BaseSuiteJsonSchemaTest.runTestFile:163 » IllegalState
[ERROR]   V4JsonSchemaTest.testIdSchemaWithUrl:287->BaseSuiteJsonSchemaTest.runTestFile:163 » IllegalState
[ERROR]   V4JsonSchemaTest.testLoadingWithId:38 » FileNotFound http://localhost:1234/sel...
[ERROR]   V4JsonSchemaTest.testRefRemoteValidator:247->BaseSuiteJsonSchemaTest.runTestFile:163 » IllegalState
[ERROR]   V6JsonSchemaTest.testRefRemoteValidator:212->BaseSuiteJsonSchemaTest.runTestFile:163 » IllegalState
[ERROR]   V7JsonSchemaTest.testRefRemoteValidator:308->BaseSuiteJsonSchemaTest.runTestFile:163 » IllegalState
[INFO]
[ERROR] Tests run: 374, Failures: 0, Errors: 6, Skipped: 56
```

src/main/java/com/networknt/schema/AnyOfValidator.java
```diff
@@ -106,6 +118,14 @@ public Set<ValidationMessage> validate(JsonNode node, JsonNode rootNode, String
                 allErrors.addAll(errors);
             }
 
+            // determine only those errors which are NOT of type "required" property missing
+            Set<ValidationMessage> childNotRequiredErrors = allErrors.stream().filter(error -> !ValidatorTypeCode.REQUIRED.getValue().equals(error.getType())).collect(Collectors.toSet());
+
+            // in case we had at least one (anyOf, i.e. any number >= 1 of) valid subschemas, we can remove all other errors about "required" properties
+            if (numberOfValidSubSchemas >= 1 && childNotRequiredErrors.isEmpty()) {
```
RenegadeWizard, line 125: Are you sure this condition is correct? If anyOf is valid (so one or more parts are present) then shouldn't we always disregard required errors? Now it seems like you're only disregarding them only if there are no other errors

AndreasALoew: My idea with this was that in case there are other errors as well, for an anyOf scenario, you'll never know which of the alternatives have been provided by intention, and which have been provided by incident, so maybe we should not fiddle with the "required" error messages.

(Note that this is different to oneOf, as in the "oneOf" case it is much more likely that the single valid subschema alternative is indeed the intended one and not valid by incident.)

But I see your point and agree that one/we can also argue to leave the second part of the condition () out...

@stevehu : Which path would you like to go?

AndreasALoew:
@stevehu: Any news such that we can make some progress here?

AndreasALoew:
@stevehu @RenegadeWizard Please tell me in case there is anything that stops you from applying my PR or at least answering my question I raised regarding @RenegadeWizard's remark:
https://github.com/networknt/json-schema-validator/pull/559#discussion_r857689540

Sorry to be a a pain in your necks here, but I really need to see some progress here: Please understand that - due to the fact that the release date in my customer project is approaching - I need this fix to be integrated into an official json-schema-validator until first week of June, or alternatively have to fork the project myself into my customer's company Git, and do the fix there...

RenegadeWizard:
Sorry i wasn't responding. I just don't understand what you are trying to achive here. The original PR was to fix unnecessary required errors in the oneOf. The condition you proposed will sift out required errors only if at least one element is present and there are no other errors, but isn't that how this validator already works? I'm sorry if my questions are dumb, but if this PR was to fix the same thing as I did with the oneOf, then I don't get how it does.

My fix was to make the required errors disappear when the element is indeed present, just not valid (in which case we still get the other errors, just leave out the required ones). What do you mean when you say "it is much more likely that the single valid subschema alternative is indeed the intended one and not valid by incident"? Can you provide some examples, cause I don't understand :disappointed: 

Sorry for being a pain in the ass, but I just want it to be okay :wink:

AndreasALoew:
@RenegadeWizard I think we still have a misunderstanding here:

Your assumption: "The condition you proposed will sift out required errors only if at least one element is present and there are no other errors, but isn't that how this validator already works?" unfortunately is **false**.

If there currently is an AnyOf validation defined that has a number of required fields in its "any of" alternatives, the **validator currently requires ALL fields marked as required anywhere across ALL alternatives to be mandatory**, and creates validation error messages that required fields are missing otherwise. This **clearly is a BUG in the present validator** and therefore must be fixed.

Now for the alternatives: I think your approach is quite straightforward and clear: if there is at least one alternative that contains values for the single/all required field(s) for this alternative, then **you propose to consider this alternative** (or rather, the first alternative in an "any of" for which the latter holds true) to be the **desired** alternative, and consequently remove all other "required field missing" validation messages for all other alternatives from the validation output.

What I am saying is: I think that the above is one valid argumentation, but:

As we here are in an "any of" and not an "one of" scenario as for your patch, by deciding to filter out **all other required errors in other alternatives**, we might as well **silently swallow** "missing required property" errors from other alternatives that the message sender would have desired to be present as well (because "any of" means that not just one, but also multiple alternatives at the same time are fine).

Let's suppose an "anyOf" with four alternatives 1 to 4, with 1AR, 1B, 1C, 2AR, 2B, 2C, 2D, 3AR, 3BR, 4AR, 4B, 4CR, 4D being the properties of each alternative, and last letter "R" means that a field is required, such as 1AR is required, 1B is optional, 2AR is required etc.

Now with your approach, if 1AR is present, the validator would unconditionally silently "forget" about any required constraints on all other reqired properties 2AR, 3AR, 3BR, 4AR, 4CR.

With my approach, if 1AR is present and 1B and 1C are also valid, the validator would assume that alternative 1 is desired and forget about the other required validation messages **if and only if there are no other validation failures in alternatives 2 to 4** than just "required" validation failures.

In case alternatives 2 to 4 would have other validation errors in addition, "my" approach would still also output validation errors of 1AR, 2AR, 3AR, 3BR, 4AR, 4CR as - in my opinion - we cannot clearly infer that alternative 1 would have been the only alternative desired to be present by the message sender.

IMO, It is really impossible to defer which of one or more "one of" alternative is desired by the message sender only from the fact that one alternative exists in which there are all "required" mandatory properties present, but other errors still exist in this or other alternatives.

Could I make the difference now clear enough, such that @stevehu , you can decide on which alternative approach (that of @RenegadeWizard  or mine) you would like to follow with regards to "swallowing" validation errors on "required" fields in other alternatives of an "any of"?

Thanks again & best regards
Andreas

RenegadeWizard:
Okay, I get it now, thanks for clearing it up ;)

PR #561, opened 2022-04-25
Prevent from throwing an exception when setting default values

josejulio:
- this happens when an expected object has a different type
   i.e. a string.
 - This is prevented by checking the type of the node before
   attemping to set the defaults


stevehu (approved):

josejulio:
> @josejulio Thanks a lot for your help.

Thank you!

PR #578, opened 2022-06-01
fixes #575

### LINKED ISSUE #575: date-time fields with valid negative offsets are marked as invalid - need to upgrade com.ethlo.time:itu to version 1.7.0
Using version 1.0.69 of json-schema-validator, date-time values with certain valid negative offsets (for example, "2022-05-25T08:27:53-02:30", Newfoundland time) fail validation.

The problem is with the com.ethlo.time:itu library. If this library is updated from version 1.5.1 to version 1.7.0, this problem goes away. We have been overriding this dependency in our applications to avoid this issue, but it would be better if json-schema-validator did the upgrade itself.

jody-mcdonnell:
This pull request upgrades com.ethlo.time:itu from version 1.5.1 to version 1.7.0. Some valid negative time zone offsets (e.g. -2:30, Newfoundland time) are marked as invalid by version 1.5.1 of the itu library.

Unit tests are included to ensure that time zone offsets validate as expected and invalid time representations are marked as such.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
         <version.mockito>2.7.21</version.mockito>
         <version.hamcrest>2.2</version.hamcrest>
         <version.undertow>2.2.14.Final</version.undertow>
-        <version.itu>1.5.1</version.itu>
+        <version.itu>1.7.0</version.itu>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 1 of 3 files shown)


stevehu:
@jody-mcdonnell Thanks a lot for your help.

Issue #582, opened 2022-06-11
Issue with unevaluatedProperties with patternProperties and type union


jkevan:
Hello, first of all thx for the great work on this library !

I discover the following issue recently using this schema:
```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "title": "Test unevaluated properties on patternProperties",
  "type": "object",
  "patternProperties": {
    "^valid_": {
      "type": ["array", "string"],
      "items": {
        "type": "string"
      }
    }
  },
  "unevaluatedProperties": false
}
```

And trying to validate this data:
```
{
  "valid_array": ["array1_value", "array2_value"],
  "valid_string": "string_value"
}
```

The validation is failing due to unevaluated properties. But it should be valid.

This is mostly due to the fact that the Type validator is collecting the evaluated properties too late:
https://github.com/networknt/json-schema-validator/blob/95925438dc1918e30c5c001de2f761de8fda30b6/src/main/java/com/networknt/schema/TypeValidator.java#L124

In current case due to Union of type we dont reach the code in charge of collecting (previous link):
https://github.com/networknt/json-schema-validator/blob/95925438dc1918e30c5c001de2f761de8fda30b6/src/main/java/com/networknt/schema/TypeValidator.java#L113
(due to the return statement).

But anyway I found it strange to make the TypeValidator collecting for evaluated properties, shouldn't be the responsibility of the PatternPropertiesValidator like the PropertiesValidator is doing ?

Disclaimer:
- we dont have the issue in case the type is using single value and not Union, so it confirms the previous analysis.
- I tested the exact same schema and data against: https://www.jsonschemavalidator.net/ and it's correctly valid

I will try to provide unit tests scenario and a fix.

Kind regards,
Kevan.

jkevan:
Hello,
Thank you for merging the PR so quickly !

Do you have an idea when next release will be ?

Best regards,
Kevan.

stevehu:
I am doing some internal testing now and will release a new version after that. Stay tuned.

jkevan:
We upgraded to latest 1.0.71 and it's all good on our side.
I'm closing the issue.

Thank you,
Best regards.

PR #583, opened 2022-06-11
Fix unevaluatedPropeties with patternProperties and type union (#582)

jkevan:
Here is a fix proposition for issue: https://github.com/networknt/json-schema-validator/issues/582

Kind regards,
Kevan.


stevehu:
@jkevan Thanks a lot for your help.

PR #585, opened 2022-06-15
Fix#584 : changed data type to preserve order of schema attributes

sabarinathan590:


stevehu:
@sabarinathan590 Thanks a lot for your help.

PR #594, opened 2022-07-31
Remove commons lang as a compile time dependency

agentgt:
This pull request remove the commons lang dependency such that it maybe excluded in maven or gradle.

There is very little usage of Commons Lang 3 in this library and most of the usage is builtin future JDK versions (e.g. `String.isBlank` is in JDK 11).

I have left the dependency in Maven for projects that accidentally depend on this project to pull Commons Lang in as a transitive dependency.

We need this for our code base as we do not allow commons lang 3 in most of our projects.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,10 +84,18 @@
             <artifactId>slf4j-api</artifactId>
             <version>${version.slf4j}</version>
         </dependency>
+        <!-- 
+        Commons lang slated for removal but kept for projects that accidentally depend on it
+        through accidental transitive runtime dependencies.
+        
+        For projects that depend on it for compile they will get an immediate failure as the
+        scope is now runtime instead of compile.
+        -->
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
             <version>${version.common-lang3}</version>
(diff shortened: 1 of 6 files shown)


stevehu (approved):

stevehu:
@agentgt. Thanks a lot for your help.

agentgt:
Thanks @stevehu !

You might want to add to the readme or whatever that if people want to exclude commons lang:

```
<dependency>
    <groupId>com.networknt</groupId>
    <artifactId>json-schema-validator</artifactId>
    <version>next version</version>
    <exclusions>
    <exclusion>
    <groupId>org.apache.commons</groupId>
    <artifactId>commons-lang3</artifactId>
    </exclusion>
    </exclusions>
</dependency>
```

Then perhaps in the next major or minor version you remove the dependency all together from the maven pom file.

stevehu:
I think we can safely remove the commons-lang3 dependency. What do you think?

agentgt:
> I think we can safely remove the commons-lang3 dependency. What do you think?

Probably. I think 99% of the time it will be ok because if someone modifies their pom to use the latest version of this library they can in theory also fix relying on the transitive dependency which is bad practice anyway.

Issue #597, opened 2022-08-05
Add Italian translation for validation messages


sbernardo:
Need to add Italian translation for messages.

PR #598, opened 2022-08-05
GH-597 - add italian translation

sbernardo:
This solves issue #597


stevehu:
@sbernardo Thanks a lot for your help.

PR #608, opened 2022-09-13
issue_563 Support adding custom message and custom validator

makeItEasyQ:
Support adding a custom message at attribute level and custom validator for all schema versions. The documented info for the  [custom validator](https://github.com/networknt/json-schema-validator/blob/master/doc/validators.md ), only supports one version, but actually, we may use different schemes by different members. So we can provide the JsonMetaSchema#getKeywords.
In this way, we can easy to add additional custom validators. For example: 
`JsonSchemaVersion jsonSchemaVersion = JsonSchemaFactory.checkVersion(versionFlag);
        JsonMetaSchema metaSchema = jsonSchemaVersion.getInstance();
        metaSchema.getKeywords().put(DATE_PATTERN, new DatePatternKeyWord(DATE_PATTERN));
        return JsonSchemaFactory.builder()
            .defaultMetaSchemaURI(metaSchema.getUri())
            .addMetaSchema(metaSchema)
            .build();`

### DIFF
--- a/src/main/java/com/networknt/schema/JsonMetaSchema.java
+++ b/src/main/java/com/networknt/schema/JsonMetaSchema.java
@@ -104,7 +104,7 @@ public class JsonMetaSchema {
         }
 
         private static Map<String, Keyword> createKeywordsMap(Map<String, Keyword> kwords, Map<String, Format> formats) {
-            final Map<String, Keyword> map = new HashMap<String, Keyword>();
+            Map<String, Keyword> map = new HashMap<String, Keyword>();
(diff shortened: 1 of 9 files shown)


makeItEasyQ:
Hi @stevehu, 
Could you please help review these changes? Thanks very much!

stevehu:
@makeItEasyQ Thanks a lot for your help. I am in the process of reviewing the code. Look good to me but I need to make sure there is no backward compatibility issue.

makeItEasyQ:
@stevehu , Thanks very much!

stevehu (approved):

PR #634, opened 2023-01-17
Quick fix for wrong custom message bug

chaosape:
The essence of the problem seems to be that keywords are unique and can only have one custom message. When there are multiple custom messages assigned to the same keyword in a schema the mutation of the custom message can result in the incorrect error messages to be output. 

This bug goes back to, at least, version 1.0.64.


stevehu:
@chaosape Thanks a lot for your help. Would you be able to update the cust-msg.md to add your use case? 

https://github.com/networknt/json-schema-validator/blob/master/doc/cust-msg.md

PR #636, opened 2023-01-24
Adding tests for overriding error messages at schema level for individual keywords

anjnerajat:
I was trying to understand the functionality implemented at networknt for supporting custom error messages with the help of tests, and realised there aren't much added here, hence created this PR. Please let me know if more cases can be covered.


stevehu:
@anjnerajat Thanks a lot for your help. The customized message is not the standard and is hard for people to follow. These test cases are really helpful.

PR #646, opened 2023-02-01
Add ability to detect spec version optionally

valfirst:
There could be cases when schemas are managed by third-parties, in this case client side logic can either use spec version defined in schema if available or some default one. Having current implementation of `SpecVersionDetector` it's not really convenient to catch exception to fallback to some default value then. Also exceptions are quite heavy mechanism, thus new method is proposed which returns an empty `Optional` value, if the schema tag is not present.

Also small refactoring of `SpecVersionDetector` class is done in order to make it true-utility class:
 - class is marked as `final`
 - private constructor is added


stevehu (approved):

stevehu:
@valfirst Thanks a lot for your help.

PR #651, opened 2023-02-06
Map BinaryNodes to type string

### LINKED ISSUE #650: Support BinaryNode
When using a Java data model with a byte[] property, json-schema-validator fails to validate because of an unknown data type.

Example:

Java model:
```java
public class Model {

      @ApiModelProperty(name = "s", value = "BASE64 encoded signature", required = true)
    private byte[] signature;
}
````
Jackson converts byte[] automatically to BASE64 strings.

JSON schema:
````json
{
  "properties": {
      "s": {
        "$id": "#/properties/s",
        "type": "string"
      }
  }
}
````

JSON file:
````json
{
  "s": "BASE64 string"
}
````

When parsing it using jackson using the Java model (with byte[] type) and then validating it, the resulting node in the jackson tree is of type BinaryTree. Validation fails then with an unknown type is not of type string error.
IMPORTANT to reproduce: Do not use jackson to directly parse into a tree. In this case the resulting jackosn node is of type TextNode and validation works fine.

k-oliver:
..to support using Java models with type byte[].
close #650

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.networknt</groupId>
     <artifactId>json-schema-validator</artifactId>
-    <version>1.0.76</version>
+    <version>1.0.77</version>
     <packaging>bundle</packaging>
     <description>A json schema validator that supports draft v4, v6, v7, v2019-09 and v2020-12</description>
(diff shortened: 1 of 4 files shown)


k-oliver:
@stevehu I added a test case for the issue. Feel free to change what is required or ask if something is missing.

stevehu (approved):

PR #626, opened 2022-12-01
Correcting the oneOf,anyOf and allOf child schema validators to use the full path

prashanthjos:
Hi @stevehu can you please have a look at this PR, I wasn't sure why we suppressed the schema path's of child schema under oneOf, anyOf, and allOf to use just the name of the respective keyword and ignore the parent path where these keywords exist. We have use cases where we update the schemas dynamically based on the complete schema path to a node. 

I have corrected these classes, can you please have a look and let us know your opinion?

If you feel the PR looks good can you please merge it to the master and release a new version?

Thank you

PR #663, opened 2023-02-16
Use full schema path to look up type validators for `anyOf` operator

### LINKED ISSUE #662: Misleading validation error for `anyOf` property when required value is invalid
### Summary

Hey everyone, we try to upgrade from version 1.0.69 to 1.0.77 and observe that the validation errors for a certain scenario changed starting with 1.0.74. The error is a bit misleading and I'd like to get your thoughts on whether the previous behavior can be restored or if it will stay this way. The detailed test case is attached. 

To sum it up, we have an object that can either be null or have a required enum property. If the property supplied is invalid, then:
* before 1.0.74, we would only get a validation error stating that `$.optionalObject.value: does not have a value in the enumeration`
* after 1.0.74, we would also get a message that `$.optionalObject: object found, null expected` which is not true, since `optionalObject` can be an object, it just happens that this object is not valid.

I'd expect to only get a message about the value in this case.

### Reproducer

```
import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.ObjectMapper;
import com.networknt.schema.JsonSchemaFactory;
import com.networknt.schema.SpecVersion;
import com.networknt.schema.ValidationMessage;
import org.junit.jupiter.api.Test;

import static org.junit.jupiter.api.Assertions.assertFalse;
import static org.junit.jupiter.api.Assertions.assertTrue;

class ReproducerTest {
    private static final String SCHEMA = """
        {
           "$schema": "http://json-schema.org/draft-07/schema#",
           "properties": {
               "optionalObject": {
                   "anyOf": [
                       {
                           "type": "null"
                       },
                       {
                           "type": "object",
                           "properties": {
                               "value": {
                                   "enum": [
                                       "one",
                                       "two"
                                   ]
                               }
                           },
                           "required": [
                               "value"
                           ]
                       }
                   ]
               }
           }
        }""";

    private static final String INVALID_VALUE = """
        {
            "optionalObject": {
                "value": "invalid"
            }
        }""";

    @Test
    void testCorrectErrorForInvalidValue() throws JsonProcessingException {
        var mapper = new ObjectMapper();
        var schema = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7).getSchema(SCHEMA);

        var errors = schema.validate(mapper.readTree(INVALID_VALUE));
        var errorMessages = errors.stream().map(ValidationMessage::getMessage).toList();

        assertTrue(errorMessages.contains("$.optionalObject.value: does not have a value in the enumeration [one, two]"));
        assertFalse(errorMessages.contains("$.optionalObject: object found, null expected"));
    }
}

```

pshevche:
### Summary
https://github.com/networknt/json-schema-validator/pull/626 changed how the schema path is set for `JsonSchema`s of `anyOf` declaration. However, it did not account for the fact that this schema path is used to identify type validators that belong to the operator. `AnyOfValidator#validate` still relied on the previous schema path and could not apply special handling to those validators (i.e. execute just the validator that matches the value). As a result, `TypeValidator`s for all options of `anyOf` property were evaluated separately and results aggregated. 

_Note_: I have no clue why the diff for `AnyOfValidator` is so ugly. The screenshot shows what exactly changed:
Screenshot 2023-02-16 at 2 14 59 PM

Resolves #662


stevehu:
@prashanthjos, Could you please review the changes to ensure they won't impact your application? Thanks.

stevehu:
@pshevche Thanks a lot for your help.

pshevche:
> @pshevche Thanks a lot for your help.

My pleasure :)

PR #666, opened 2023-03-01
Clarify usage of Apache commons lang in README.md

loadedice:
Much like in https://github.com/networknt/json-schema-validator/pull/594 apache commons is banned in a code base I work in. I assumed this library was not suitable for use based on the README but after searching the issues I found the aforementioned pull request. 

Personally I believe it  should be fine to just remove it from `pom.xml` and update the README.md accordingly; as it has been several months since that change was merged, but acknowledge it may break _some_ users functionality if they are not careful about accidental transitive dependency use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,12 @@ Here are the dependencies.
     <artifactId>slf4j-api</artifactId>
     <version>${version.slf4j}</version>
 </dependency>
-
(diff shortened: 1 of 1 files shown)


stevehu:
@loadedice Thanks a lot for your help.

PR #671, opened 2023-03-06
Support time offsets in `format: time`

### LINKED ISSUE #669: `time` validation incomplete
I have a schema using `draft-07`. It has a property using `format: time`. When validating the following input `07:00:00-06:00`, `json-schema-validator` 1.0.72 generates an error `X: does not match the time pattern ^\d{2}:\d{2}:\d{2}(?:\.\d+)?$`. However, I believe my input to be valid per the [spec](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-7.3.1). It appears the `time-offset` may not be implemented.

JDziurlaj:
Resolves #669. Note that the regex has changed from what was originally agreed upon in #669. This is in order to reuse most of the test suite that was developed for #575.

### DIFF
--- a/src/main/java/com/networknt/schema/JsonMetaSchema.java
+++ b/src/main/java/com/networknt/schema/JsonMetaSchema.java
@@ -40,7 +40,7 @@ public class JsonMetaSchema {
 
     // this section contains formats that is common for all specification versions.
     static {
-        COMMON_BUILTIN_FORMATS.add(pattern("time", "^\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?$"));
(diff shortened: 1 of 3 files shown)


stevehu (approved):

stevehu:
@JDziurlaj Thanks a lot for the enhancement and test cases to cover all the scenarios. The CI build complains about one test. Could you please take a look? 

```
Error:  Failures: 
Error:    JsonMetaSchemaTest.timeFormatValidation:21 expected: <true> but was: <false>
[INFO] 
Error:  Tests run: 513, Failures: 1, Errors: 0, Skipped: 75
[INFO] 
```

PR #691, opened 2023-04-06
Uses JUnit dynamic tests to generate tests from specification files

### LINKED ISSUE #690: Refactor unit-tests from JSON Schema Test Suite
JSON Schema Test Suite has several tests for each test-case and there is one or more test-cases in each specification file. Currently, this project treats each specification file as a single unit-test. As a result, a single test failure causes the unit-test to fail and no subsequent tests are performed. If the first test in a specification file fails, for example, no other tests are executed.

Several specification files are currently disabled simply because a single test fails. For example, content.json has four test-cases with a total of 18 tests. The unit-test is disabled so none of the tests are executed.

I propose refactoring the unit-tests to use JUnit Jupiter's dynamic test facility to create a separate unit-test for each test in a test-case (and each test-case in a specification file). This would provided several benefits:
- Better test coverage since more tests are actually executed
- Better reporting since each test is a single line-item in the report
- Better reporting of the issue found (e.g., each test can report more detailed information than "expected true but was false")
- Easier to investigate the cause of some obscure bugs such as #285, #313 and #587
- More fine-grained control over which tests are excluded (i.e., instead of excluding a whole specification file, support excluding a specific test-case or individual test)

**_The specification files also need refreshing but that is best left to a separate ticket._**

This is an example of how this would appear when executed from the command line.
```
[INFO] +--JSON Schema Test Suite - 1.063s
[INFO] |  +-- [OK] validation of duration strings (src/test/resources/draft2019-09/optional/format/duration.json) Draft 2019-09 no elements present - 0.001s
[INFO] |  +-- [XX] validation of duration strings (src/test/resources/draft2019-09/optional/format/duration.json) Draft 2019-09 no time elements present - 0s
[INFO] |  +-- [OK] validation of duration strings (src/test/resources/draft2019-09/optional/format/duration.json) Draft 2019-09 no date or time elements present - 0s
```

This is how it appears in Eclipse (IntelliJ has similar views).
image
image

fdutton:
Resolves #690 

**NOTE**: I reverted the Logback version change. Version 1.4.6 was built with Java 11.


stevehu (approved):

stevehu:
@fdutton Thanks a lot for your help. We still have so many Java 8 users and hope they can upgrade to Java 11 or Java 17.

Issue #587, opened 2022-07-01
V2020-12 not validating items: { "type" } correctly


rustermi:
I have tests build around this validation library and noticed a type error since I migrated to V2020-12. Given a schema with the following property:
```json
"numberArray": {
  "type": "array",
  "items": {
    "type": "number"
  }
}
```

The following object should pass `{"numberArray": [1, 2, 3]}` and this one should not: `{"numberArray": [1, "wrong type", 3]}`. Since the upgrade to the new draft version, both objects are validated as error-free.

The original error only appears if the schema is modified as follows:
```json
"numberArray": {
  "type": "array",
  "prefixItems": [{}],
  "items": {
    "type": "number"
  }
}
```

However, the spec features [this specific example without any `prefixItems`](https://json-schema.org/understanding-json-schema/reference/array.html#items):
<img width="790" alt="items" src="https://user-images.githubusercontent.com/97174955/176888452-b5d3d374-eb41-4577-8b00-5a4da9263c89.png">

@open-abbott I ping you, first to say thank you for the implementation and second because maybe you quickly know what to do 😃

rustermi:
And by the way, as `items` is superseded by `prefixItems` in this implementation of V2020-12, I also get this warning:

```
com.networknt.schema.JsonMetaSchema - Unknown keyword items - you should define your own Meta Schema. If the keyword is irrelevant for validation, just use a NonValidationKeyword
```
because it is not added to the `JsonMetaSchema` of the `Version202012` class.

zeldigas:
Enum is also not validated for arrays

open-abbott:
Hey @rustermi !  Currently swamped with other projects, but I'm guessing that the problem lies with the fact that `items` was semantically redefined in the new spec, so there probably needs to be a new `ItemsValidator` written that behaves correctly, then apply that to the 20201220 version and onward in the `ValidatorTypeCode`.  Cheers!

mrozycki-tink:
Had the same issue with items not being validated in an array with V2020-12. The `items` keyword is configured in `ValidatorTypeCode` to only be supported up to V2019-09, so that part of the schema doesn't even generate a validator.

Our workaround was to revert to V2019-09 for now, but it would be great to be able to go back to V2020-12 at some point.

stevehu:
In 2020-12 release, the items and additionalItems keywords have been replaced with prefixItems and items. They are trying to reuse the keywords with different meanings. We are open to all suggestions on the design to resolve the issue.

[...]

PR #710, opened 2023-04-16
Adds support for the Draft 2020-12 interpretation of `prefixItems` and `items`.

fdutton:


stevehu:
@fdutton, you have created a specific validator for the 2020-12 version for the item keyword, which is redefined for a different meaning. There are some discussions about handling this situation and the pros and cons in https://github.com/networknt/json-schema-validator/issues/587

I was thinking of using the same validators but making these validators version aware. As we cannot predict the schema spec team if they are going to give the same keyword other meanings in the future or revert back to the old meanings, it might be safer to build the version-aware code in the same validator. This will make the validator more complicated but still manageable. The other benefit is to handle parent schema and subschema using different versions. If we pass the version and identify the new version when the schema is loaded in the context, all validators should know what is the current version and will react accordingly. What is your idea?

fdutton:
@stevehu I made ContainsValidator and those that accept 1.0 as an integer version aware but it gets tricky. Some of the validators can exist on child schemas and one has to walk up the schema to know which version to use. I think there may be some undiscovered issues in this functionality that will become apparent once I get all the test-cases working.

My thought in this case was that the factory was in the best position to know what version applies and select the right validators. I think this is a better approach but requires supporting multiple validators for the same keyword based on the dialect. It will require some significant refactoring to support custom meta-schemas and vocabularies so spreading the version detection logic throughout the validators may become problematic.

stevehu:
@fdutton You have convinced me. Let's make it work first, and we can always refactor it later if we need to support dynamic version detection for subschemas. Thanks.

PR #786, opened 2023-05-25
Allow walking of schema for items keyword when non-array node is provided

### LINKED ISSUE #785: ItemsValidator202012 doesn't walk for non-array nodes
ItemsValidator202012 unlike ItemsValidator doesn't walk for non-array nodes. This blocks simply walking the schema which contains items keyword, with empty data node with shouldValidateSchema = false

anjnerajat:
Resolves #785 

Fixing ItemsValidator202012 walk flow.
Making UnevaluatedProperties CollectorContext's key public as was in 1.0.76.


stevehu (approved):

stevehu:
@anjnerajat Thanks for the quick fix.

Issue #844, opened 2023-07-31
I18N Custom Locale or Can we support more languages (i.e spanish, Japanese, Korean )


channaveer1:
Hi Team, 

I want to support more languages in i18N as ResourceBundle has few "jsv-messages", Please let me know any possibility of overriding or how to add more languages. or any chances of contributing from our end.

Thanks
Channaveer patil
yakshya10@gmail.com

stevehu:
You can submit PRs to add more translations. Most team members are English speakers, and it is hard for us to help.

channaveer1:
@stevehu you want me to give translated messages in other languages? so that can be added to ResourceBundles

stevehu:
Please submit a PR with your resource bundles. Thanks.

channaveer1:
Okay i will raise soon.

channaveer1:
Hi @stevehu i need permission so that  i can raise PR to json-schema-validator, please find the error in below image and suggest what to do next. or provide the access to contribute.  

![image](https://github.com/networknt/json-schema-validator/assets/16414370/9eaeb395-1b3e-47df-a7c8-f6ed3d6a135c)

costas80:
Hi @channaveer1 , note that you can also supply your own resource bundle as part of the validation's configuration (see the docs via the README page). Of course if you do this you need to make sure your bundle aligns in terms of keys with the default one, but at least like that you have full flexibility to e.g. override messages or support other locales.

stevehu:
More locales are supported in the latest version. If anyone still has specific issues, please open a new issue. Thanks.

PR #852, opened 2023-08-18
New resource bundle languages added for issue  #844

channaveer1:
I have raised request for new languages for error messages #I18N


stevehu:
@channaveer1 Thanks a lot for the effort. I saw some files contain question marks only. Is it due to the editor that cannot support UTF encoding?

channaveer1:
Yes i believe i have translated correctly. may be IDE problem

PR #866, opened 2023-10-06
Fix identation in example in walkers.md

bpaquet:


stevehu (approved):

stevehu:
@bpaquet Thanks for your help.

PR #883, opened 2023-10-24
docs: clarify commons-lang3 exclusion only required for 1.0.81 and older

JonasGroeger:
Since its the only entry, the file https://github.com/networknt/json-schema-validator/blob/master/NOTICE could probably also be deleted.


stevehu (approved):

JonasGroeger:
Hey @stevehu,

could you check out my two remarks on the PR description? What do you think?

PR #886, opened 2023-10-26
Fix invalid class passed to getLogger

brempusz:
getLogger parameter should match class name, where the logger is declared.


stevehu:
@brempusz Thanks a lot for your help.

PR #887, opened 2023-10-27
Fix pl_PL message translations

brempusz:
I noticed there is a lot of errors in other languages (`?` or `�` characters), but I will not able to fix those.

I used `utf-8` when editing properties in IDEA. See https://openjdk.org/jeps/226


stevehu (approved):

stevehu:
@brempusz Thanks a lot for your help.

Issue #872, opened 2023-10-11
Please revert fix from #815 (pattern validation)


kool79:
Please revert fix made in scope of #815.
This fix was deployed based on incorrect assumption that java works incorrectly with regexp.
But actually issue is not in java. Issue (reported in #814) was caused by incorrect regexp which author uses us json schema.

STR:
GIVEN:
user (in schema) defines regexp as:   `^red|banana$`   (starts with 'red' OR ends with 'banana')
to pass any strings like: "red", "banana", "red apple",  "green banana", "red oval apple"

But the #815 fix will will change regexp to `^(red|banana)$` which will pass ONLY "red" or "banana"

For more details please see my comment to the code
https://github.com/networknt/json-schema-validator/pull/815/files#diff-776968543c85221a4907857a99fb117ab32872c5f0c020a717c11fe8162ef310R12-R24

kool79:
Alternative tool for visualization: https://regexper.com/#%5Ered%7Cbanana%24
![scr_2023_10_11T15_39_37](https://github.com/networknt/json-schema-validator/assets/4625317/b16ad81a-ceb5-43fb-bdfd-4424a6eccbcb)
![scr_2023_10_11T15_39_56](https://github.com/networknt/json-schema-validator/assets/4625317/391e82ad-b050-48f7-b775-64d9ff09ce56)

stevehu:
@fdutton What do you think? Thanks.

Stephan202:
I started looking into this, and the topic is quite a rabbit hole, both in terms of regex matching behavior and invalidate test cases. I tried my best at cleaning things up in #888. @kool79 @fdutton reviews very welcome!

PR #888, opened 2023-11-03
Fix JDK regex support

Stephan202:
Summary of changes:
- Fix the test resources introduced by #783 by moving the `regex` fields, such that the test framework does not skip them with a "Not a valid test case" message.
- Revert the changes introduced by #815, as those are simply incorrect.
- Extend the test coverage introduced by #815 by (a) updating the test regexes to match their intended semantics and (b) include a few negative test cases.
- Partially revert the change introduced by #783: the use of `Matcher#find()` is correct, but the `hasStartAnchor` and `hasEndAnchor` logic introduces more bugs than the issue it aims to solve.
- Extend the test coverage introduced by #783, by introducing regexes that are not covered by the `hasStartAnchor`/`hasEndAnchor` logic.
- Update the Joni regular expression integration such that it passes more of the test cases.
- Disable the "trailing newline" test cases, as these are currently not handled correctly by either regex implementation.

### DIFF
--- a/src/main/java/com/networknt/schema/regex/JDKRegularExpression.java
+++ b/src/main/java/com/networknt/schema/regex/JDKRegularExpression.java
@@ -1,33 +1,16 @@
 package com.networknt.schema.regex;
 
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 class JDKRegularExpression implements RegularExpression {
     private final Pattern pattern;
-    private final boolean hasStartAnchor;
-    private final boolean hasEndAnchor;
 
     JDKRegularExpression(String regex) {
-        // The patterns in JSON Schema are not implicitly anchored so we must
-        // use Matcher.find(). However, this method does not honor the end
-        // anchor when immediately preceded by a quantifier (e.g., ?, *, +).
-        // To make this work in all cases, we wrap the pattern in a group. 
-        this.hasStartAnchor = '^' == regex.charAt(0);
-        this.hasEndAnchor = '$' == regex.charAt(regex.length() - 1);
-        String pattern = regex;
-        if (this.hasEndAnchor) {
-            pattern = pattern.substring(this.hasStartAnchor ? 1 : 0, pattern.length() - 1);
-            pattern = '(' + pattern + ")$";
-            if (this.hasStartAnchor) pattern = '^' + pattern;
-        }
-        this.pattern = Pattern.compile(pattern);
+        this.pattern = Pattern.compile(regex);
     }
 
     @Override
     public boolean matches(String value) {
-        Matcher matcher = this.pattern.matcher(value);
-        return matcher.find() && (!this.hasStartAnchor || 0 == matcher.start()) && (!this.hasEndAnchor || matcher.end() == value.length());
+        return this.pattern.matcher(value).find();
     }
-
 }
--- a/src/main/java/com/networknt/schema/regex/JoniRegularExpression.java
+++ b/src/main/java/com/networknt/schema/regex/JoniRegularExpression.java
@@ -21,7 +21,7 @@ class JoniRegularExpression implements RegularExpression {
             .replace("\\S", "[^ \\f\\n\\r\\t\\v\\u00a0\\u1680\\u2000-\\u200a\\u2028\\u2029\\u202f\\u205f\\u3000\\ufeff]");
 
         byte[] bytes = s.getBytes(StandardCharsets.UTF_8);
-        this.pattern = new Regex(bytes, 0, bytes.length, Option.NONE, UTF8Encoding.INSTANCE, Syntax.ECMAScript);
+        this.pattern = new Regex(bytes, 0, bytes.length, Option.SINGLELINE, UTF8Encoding.INSTANCE, Syntax.ECMAScript);
     }
 
     @Override
--- a/src/test/java/com/networknt/schema/regex/Issue814Test.java
+++ b/src/test/java/com/networknt/schema/regex/Issue814Test.java
@@ -8,40 +8,62 @@ class Issue814Test {
 
     @Test
     void jdkTypePattern() {
-        JDKRegularExpression ex = new JDKRegularExpression("^list|date|time|string|enum|int|double|long|boolean|number$");
+        JDKRegularExpression ex = new JDKRegularExpression("^(list|date|time|string|enum|int|double|long|boolean|number)$");
         assertTrue(ex.matches("list"));
         assertTrue(ex.matches("string"));
         assertTrue(ex.matches("boolean"));
         assertTrue(ex.matches("number"));
         assertTrue(ex.matches("enum"));
+        assertFalse(ex.matches("listZ"));
+        assertFalse(ex.matches("AenumZ"));
+        assertFalse(ex.matches("Anumber"));
     }
 
     @Test
     void jdkOptionsPattern() {
-        JDKRegularExpression ex = new JDKRegularExpression("^\\d*|[a-zA-Z_]+$");
-        assertTrue(ex.matches("external"));
-        assertTrue(ex.matches("external_gte"));
-        assertTrue(ex.matches("force"));
-        assertTrue(ex.matches("internal"));
+        JDKRegularExpression ex = new JDKRegularExpression("^\\d|[a-zA-Z_]$");
+        assertTrue(ex.matches("5"));
+        assertTrue(ex.matches("55"));
+        assertTrue(ex.matches("5%"));
+        assertTrue(ex.matches("a"));
+        assertTrue(ex.matches("aa"));
+        assertTrue(ex.matches("%a"));
+        assertTrue(ex.matches("%_"));
+        assertTrue(ex.matches("55aa"));
+        assertTrue(ex.matches("5%%a"));
+        assertFalse(ex.matches(""));
+        assertFalse(ex.matches("%"));
+        assertFalse(ex.matches("a5"));
     }
 
     @Test
     void joniTypePattern() {
-        JoniRegularExpression ex = new JoniRegularExpression("^list|date|time|string|enum|int|double|long|boolean|number$");
+        JoniRegularExpression ex = new JoniRegularExpression("^(list|date|time|string|enum|int|double|long|boolean|number)$");
         assertTrue(ex.matches("list"));
         assertTrue(ex.matches("string"));
         assertTrue(ex.matches("boolean"));
         assertTrue(ex.matches("number"));
         assertTrue(ex.matches("enum"));
+        assertFalse(ex.matches("listZ"));
+        assertFalse(ex.matches("AenumZ"));
+        assertFalse(ex.matches("Anumber"));
     }
 
     @Test
     void joniOptionsPattern() {
-        JoniRegularExpression ex = new JoniRegularExpression("^\\d*|[a-zA-Z_]+$");
-        assertTrue(ex.matches("internal"));
-        assertTrue(ex.matches("external"));
-        assertTrue(ex.matches("external_gte"));
-        assertTrue(ex.matches("force"));
+        JoniRegularExpression ex = new JoniRegularExpression("^\\d|[a-zA-Z_]$");
+        assertTrue(ex.matches("5"));
+        assertTrue(ex.matches("55"));
+        assertTrue(ex.matches("5%"));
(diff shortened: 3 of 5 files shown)


src/main/java/com/networknt/schema/regex/JDKRegularExpression.java
```diff
@@ -1,33 +1,16 @@
...
-        }
-        this.pattern = Pattern.compile(pattern);
+        this.pattern = Pattern.compile(regex);
     }
 
     @Override
     public boolean matches(String value) {
-        Matcher matcher = this.pattern.matcher(value);
-        return matcher.find() && (!this.hasStartAnchor || 0 == matcher.start()) && (!this.hasEndAnchor || matcher.end() == value.length());
+        return this.pattern.matcher(value).find();
```
Stephan202, line 14: This matches the original suggestion by @tvahrst in #782, and seems to be closest to what the spec intents (modulo the newline case).

src/main/java/com/networknt/schema/regex/JoniRegularExpression.java
```diff
@@ -21,7 +21,7 @@ class JoniRegularExpression implements RegularExpression {
             .replace("\\S", "[^ \\f\\n\\r\\t\\v\\u00a0\\u1680\\u2000-\\u200a\\u2028\\u2029\\u202f\\u205f\\u3000\\ufeff]");
 
         byte[] bytes = s.getBytes(StandardCharsets.UTF_8);
-        this.pattern = new Regex(bytes, 0, bytes.length, Option.NONE, UTF8Encoding.INSTANCE, Syntax.ECMAScript);
+        this.pattern = new Regex(bytes, 0, bytes.length, Option.SINGLELINE, UTF8Encoding.INSTANCE, Syntax.ECMAScript);
```
Stephan202, line 24: I did not find the Joni documentation, but this change makes more test cases pass; please carefully review.

src/test/java/com/networknt/schema/regex/Issue814Test.java
```diff
@@ -8,40 +8,50 @@ class Issue814Test {
 
     @Test
     void jdkTypePattern() {
-        JDKRegularExpression ex = new JDKRegularExpression("^list|date|time|string|enum|int|double|long|boolean|number$");
+        JDKRegularExpression ex = new JDKRegularExpression("^(list|date|time|string|enum|int|double|long|boolean|number)$");
```
Stephan202, line 11: As mentioned by @kool79 in #872, if a user wishes only to match these words, then the parentheses are required.

src/test/java/com/networknt/schema/regex/Issue814Test.java
```diff
@@ -8,40 +8,50 @@ class Issue814Test {
...
         assertTrue(ex.matches("enum"));
+        assertFalse(ex.matches("listZ"));
+        assertFalse(ex.matches("AenumZ"));
+        assertFalse(ex.matches("Anumber"));
     }
 
     @Test
     void jdkOptionsPattern() {
-        JDKRegularExpression ex = new JDKRegularExpression("^\\d*|[a-zA-Z_]+$");
+        JDKRegularExpression ex = new JDKRegularExpression("^\\d+|[a-zA-Z_]+$");
```
Stephan202, line 24: Note that I changed `*` to `+`: because `^\d*` is vacuously true: it matches the empty string at the start of any input. (Without this the `internal9` test case below would also yield `true`.)

src/test/resources/draft2020-12/issue495.json
```diff
@@ -1,61 +1,105 @@
 [
     {
         "description": "issue495 using ECMA-262",
-        "regex": "ecma-262",
```
Stephan202, line 4: I moved this fields, because `regex` is defined in the `TestSpec` class, not inside `TestCase`. Without this the tests were skipped...

src/test/resources/draft2020-12/issue495.json
```diff
@@ -1,61 +1,105 @@
...
         "schema": {
             "$schema": "https://json-schema.org/draft/2020-12/schema",
-            "pattern": "^[a-z]{1,10}$",
+            "patternProperties": {
+              "^[a-z]{1,10}$": true,
+              "(^1$)": true
+            },
```
Stephan202, line 9: The `pattern` keyword didn't seem to work, so I switched to `patternProperties`; let me know if something else was intended. Also added an extra test case to show why inspecting the regex will not work in general.

src/test/resources/draft2020-12/issue495.json
```diff
@@ -1,61 +1,105 @@
...
             {
                 "description": "trailing newline",
+                "regex": "jdk",
                 "data": { "aaa\n": 3 },
-                "valid": false
+                "valid": false,
+                "disabled": true,
+                "comment": "Test fails"
+            },
```
Stephan202, line 85: This is the only test case on which this PR regresses. Both regex implementation suffer from the same issue:
```
jshell> Pattern.compile("^a$").matcher("a\n").find()
$1 ==> true
jshell> Pattern.compile("(^a$)").matcher("a\n").find()
$2 ==> true
```

Stephan202 (commented):
Added some context. Still thinking about the "trailing newline" cases, but didn't find any solution yet.

This PR resolves #872.

Suggested commit message:
```
Fix JDK regex support

Summary of changes:
- Fix the test resources introduced by #783 by moving the `regex`
  fields, such that the test framework does not skip them with a "Not a
  valid test case" message.
- Revert the changes introduced by #815, as those are simply incorrect.
- Extend the test coverage introduced by #815 by (a) updating the test
  regexes to match their intended semantics and (b) include a few 
  negative test cases.
- Partially revert the change introduced by #783: the use of
  `Matcher#find()` is correct, but the `hasStartAnchor` and 
  `hasEndAnchor` logic introduces more bugs than the issue it aims to
  solve.
- Extend the test coverage introduced by #783, by introducing regexes
  that are not covered by the `hasStartAnchor`/`hasEndAnchor` logic.
- Update the Joni regular expression integration such that it passes
  more of the test cases.
- Disable the "trailing newline" test cases, as these are currently not 
  handled correctly by either regex implementation.

Resolves #872.
```

src/test/resources/draft2020-12/issue495.json
Stephan202: Okay, I found why this happens. [The documentation states](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/regex/Pattern.html#lt):
>  The regular expression `$` matches at the end of the entire input sequence, but also matches just before the last line terminator if this is not followed by any other input character. Other line terminators are ignored, including the last one if it is followed by other input characters.

So _in theory_ we could avoid this by transforming all `$` anchors by prepending a negative lookahead, like so:
```
jshell> Pattern.compile("^a(?!\n|\r\n|\r|\\u0085|\\u2028|\\u2029)$").matcher("a\n").find()
$1 ==> false
```

But that will require careful parsing of the input regex, because the `$` could be escaped, or inside a `\Q`...`\E` section, or... Quite frankly, that way lies madness; would like to keep it out of this PR.

src/test/java/com/networknt/schema/regex/Issue814Test.java
```diff
@@ -8,40 +8,50 @@ class Issue814Test {
...
     }
 
     @Test
     void joniOptionsPattern() {
-        JoniRegularExpression ex = new JoniRegularExpression("^\\d*|[a-zA-Z_]+$");
+        JoniRegularExpression ex = new JoniRegularExpression("^\\d+|[a-zA-Z_]+$");
+        assertTrue(ex.matches("0external"));
         assertTrue(ex.matches("internal"));
         assertTrue(ex.matches("external"));
         assertTrue(ex.matches("external_gte"));
```
kool79, line 52: I think that similar examples can be removed, and the next can be added:
(matches left part: starts with digit:)
assertTrue - "5"
assertTrue - "55"
assertTrue - "5%"
(matches right part: ends with alpha-or-undescore:)
assertTrue - "a"
assertTrue - "aa"
assertTrue - "%a"
assertTrue - "%_"
(matches both parts)
assertTrue - "55aa"
assertTrue - "5%%a"

assertFalse: ""
assertFalse: "%"
assertFalse: "a5"

src/test/java/com/networknt/schema/regex/Issue814Test.java
```diff
@@ -8,40 +8,50 @@ class Issue814Test {
...
         assertTrue(ex.matches("force"));
+        assertFalse(ex.matches("internal9"));
```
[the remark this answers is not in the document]
Stephan202: Right you are!

src/test/java/com/networknt/schema/regex/Issue814Test.java
Stephan202: Sure, lemme replace them.

Stephan202 (commented):
Thanks for taking a look @kool79!

I squashed the following changes into the commit:
```diff
diff --git a/src/test/java/com/networknt/schema/regex/Issue814Test.java b/src/test/java/com/networknt/schema/regex/Issue814Test.java
index 3e7d42d..15ab31c 100644
--- a/src/test/java/com/networknt/schema/regex/Issue814Test.java
+++ b/src/test/java/com/networknt/schema/regex/Issue814Test.java
@@ -21,13 +21,19 @@ class Issue814Test {
 
     @Test
     void jdkOptionsPattern() {
-        JDKRegularExpression ex = new JDKRegularExpression("^\\d+|[a-zA-Z_]+$");
-        assertTrue(ex.matches("0external"));
-        assertTrue(ex.matches("external"));
-        assertTrue(ex.matches("external_gte"));
-        assertTrue(ex.matches("force"));
-        assertTrue(ex.matches("internal"));
-        assertFalse(ex.matches("internal9"));
+        JDKRegularExpression ex = new JDKRegularExpression("^\\d|[a-zA-Z_]$");
+        assertTrue(ex.matches("5"));
+        assertTrue(ex.matches("55"));
+        assertTrue(ex.matches("5%"));
+        assertTrue(ex.matches("a"));
+        assertTrue(ex.matches("aa"));
+        assertTrue(ex.matches("%a"));
+        assertTrue(ex.matches("%_"));
+        assertTrue(ex.matches("55aa"));
+        assertTrue(ex.matches("5%%a"));
+        assertFalse(ex.matches(""));
+        assertFalse(ex.matches("%"));
+        assertFalse(ex.matches("a5"));
     }
 
     @Test
@@ -45,13 +51,19 @@ class Issue814Test {
 
     @Test
     void joniOptionsPattern() {
-        JoniRegularExpression ex = new JoniRegularExpression("^\\d+|[a-zA-Z_]+$");
-        assertTrue(ex.matches("0external"));
-        assertTrue(ex.matches("internal"));
-        assertTrue(ex.matches("external"));
-        assertTrue(ex.matches("external_gte"));
-        assertTrue(ex.matches("force"));
-        assertFalse(ex.matches("internal9"));
+        JoniRegularExpression ex = new JoniRegularExpression("^\\d|[a-zA-Z_]$");
+        assertTrue(ex.matches("5"));
+        assertTrue(ex.matches("55"));
+        assertTrue(ex.matches("5%"));
+        assertTrue(ex.matches("a"));
+        assertTrue(ex.matches("aa"));
+        assertTrue(ex.matches("%a"));
+        assertTrue(ex.matches("%_"));
+        assertTrue(ex.matches("55aa"));
+        assertTrue(ex.matches("5%%a"));
+        assertFalse(ex.matches(""));
+        assertFalse(ex.matches("%"));
+        assertFalse(ex.matches("a5"));
     }
 
 }
```

(I squashed, because looking at the `master` branch, it appears that the maintainer of this repository uses GitHub's squash-and-rebase feature without cleaning up the commit message auto-generated by GitHub, which creates a rather messy result if there's more than one commit.)

stevehu:
@Stephan202 @kool79 Thanks a lot for the hard work. Do you think we are ready to merge? @fdutton Do you have any concerns?

Stephan202:
@stevehu I think that not supporting the trailing newline case is an acceptable compromise given the larger context; I'm not planning other changes at this time :+1:

(We should trigger the build workflow, though, to make sure that the tests also pass on GHA.)

Stephan202:
@stevehu yes, I feel comfortable doing so :+1: (But don't have permission to ;).)

PR #896, opened 2023-11-23
Refactor to remove ThreadLocal usage

### LINKED ISSUE #849: CollectorContext is empty in `walk` return value
Having `isResetCollectorContext=true` is a good way to keep collector context "pure" between runs. But if you need to get collector context, the only option you have is to set `isResetCollectorContext` to `false` and then reset the context manually after you're done. It is so because of [this code](https://github.com/networknt/json-schema-validator/blob/3ff32ee3fbb1c6287c6dfeda14a8bdeb965ef6c3/src/main/java/com/networknt/schema/JsonSchema.java#L518-L523) - while the library copies collector context into `ValidationResult`, it keeps its reference and when execution reaches `finally` block it will reset the context in both returned value and actual "current" state. `CollectorContext` [creates new objects on reset](https://github.com/networknt/json-schema-validator/blob/3ff32ee3fbb1c6287c6dfeda14a8bdeb965ef6c3/src/main/java/com/networknt/schema/CollectorContext.java#L219-L220), but as soon as we're keeping the reference to the same `CollectorContext` object in `ValidationResult` it will be reset everywhere.

I don't know if it was done intentionally, but it looks like a bug for me and probably `ValidationResult` should have a copy of `CollectorContext`. Otherwise you have to do something like this:

```java
try {
    ValidationResult result = schema.walk(object, true);
    // do something here
} finally {
    schema.getCollectorContext().reset();
}
```

i.e. sort of replicate the code that the library already has. Additionally, you can easily forget to reset it in some places (if you do multiple operations with the schema in different functions, and `isResetCollectorContext` is assigned at the schema creation so you have to either use multiple schemas that are different in this flag only or somehow make sure that you always reset the context).

### LINKED ISSUE #894: CollectorContext stored in ThreadLocal is causing ClassLoader leak
We are using this library for json schema validation with hazelcast streaming (jet). Hazelcast creates classloader on demand for the pipeline execution. The classloader then loads the classes doing schema validation. The classloader is then cleaned up when the pipeline is cancelled.
But reference to classloader is not GCed because the thread local inside ThreadInfo class has reference to collector context which is never removed thread local so even through the hazelcast framework is not using this class loader anymore the classloader is never eligible for GC.
This end up causing accumulation of these classloaders and classes referenced by these classloaders and our Metaspace grows over time and we ran out of memory.

Path from GC root.
image

justin-tay:
Resolves #894, resolves #849

Removes the use of `ThreadLocal` to store context and explicitly passes the context as a parameter where needed.

The use of a `ThreadLocal` tends to be problematic. This is particularly the case when a reactive stack is used where the caller might have no control over the executing thread and it is difficult to tell when it should be cleared.

This follows the principle of least surprise. The caller needs to explicitly reuse the context for instance to collect information across runs.

These changes are unfortunately a breaking API change but I think it is preferable to working around the issues that arise due to the use of a `ThreadLocal`.

The following are the main API changes, typically to accept an `ExecutionContext` as a parameter
* `com.networknt.schema.JsonSchema`
* `com.networknt.schema.JsonValidator`
* `com.networknt.schema.Format`
* `com.networknt.schema.walk.JsonSchemaWalker`
* `com.networknt.schema.walk.WalkEvent`

`JsonSchema` was modifier to optionally accept an `ExecutionContext` for the `validate`, `validateAndCollect` and `walk` methods. For methods where no `ExecutionContext` is supplied, one is created for each run in the `createExecutionContext` method. 

`ValidationResult` was modified to store the `ExecutionContext` of the run which is also a means of reusing the context, by passing this context information from the `ValidationResult` to following runs.


stevehu:
@fdutton @prashanthjos I think the idea is great. What do you think? Is there any impact for existing codebase?

timocov:
This looks fantastic to me. Less working with a global state - better.

rphilipp64:
This is ([again](/networknt/json-schema-validator/pull/552#discussion_r861899259)) a **breaking change** that requires at least a bump of the minor version. (In this case, I'd argue even a major version bump). Again I find myself having to look at this repository to find the PR that introduced this, after an automated patch(⚠️) version bump PR build failed.

Sadly, there is no clear description in the [changelog](/networknt/json-schema-validator/releases/tag/1.0.88) or this PR description on what I need to do as integrator. (How do I create/get an `ExecutionContext`)? An upgrade mention in the changelog or this PR description would have been nice. The PR description even mentions this is breaking. Why was the version not bumped accordingly?

I realize this comment may come across a bit harsh, it's not meant that way. I very much appreciate this component and the work you put into this. But there's a reason a convention such as [semver[(https://semver.org/) exists. Your integrators need to be aware when they may need to pay extra attention/do extra work after a version bump.

stevehu:
@rphilipp64 You are right. I will remove the latest release and bump up the minor version. Thanks a lot for pointing it out.

stevehu:
1.1.0 is released and 1.0.88 is removed. Thanks.

Issue #665, opened 2023-02-24
Can't load JSON schemas with URN value in id field (draft V4)


Non0mnisMoriar:
We are designing our own JSON schemas using URNs as unique _id_ values (such as **"id": "urn:foo:bar:object"**), as is supposed to be compliant with the JSON schema definition. We're using (and stuck to) the Draft V4 version for compatibility with other components reasons.
I am trying to load the JSON schema file into a JsonSchema object using this simple function:

```
public static JsonSchema getJsonSchemaFromClasspath(String name) {
    JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
    InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream(name);
    return factory.getSchema(is);
}
```
(_name_ argument value being here "json_schema/my_foobar_schema.json" for instance)

It does not work though, the error thrown being:

```
com.networknt.schema.JsonSchemaException: urn:foo:bar:object: null is an invalid segment for URI {2}
	at com.networknt.schema.JsonSchema.combineCurrentUriWithIds(JsonSchema.java:112) ~[json-schema-validator-1.0.77.jar:?]
```
When trying to debug what was happening, I figured out the above exception message above is thrown after catching another exception happening in the code, which is the actual cause of the error:

```
java.lang.IllegalArgumentException: Unsupported URI scheme encountered: urn
	at com.networknt.schema.uri.URISchemeFactory.getFactory(URISchemeFactory.java:55)
```
Meaning that URNs might not be supported at all?

I found a similar comment by @Arno538 on issue 100: https://github.com/networknt/json-schema-validator/issues/100#issuecomment-614755074, and therefore tried to revert back to version 1.0.29 since it was apparently working for them up to that version. It still wouldn't work though, the same "Unsupported URI scheme encountered: urn" exception appearing during debugging, with the only difference being the actual exception thrown being an NPE:

```
java.lang.NullPointerException: Cannot invoke "java.net.URI.toString()" because "currentUri" is null
	at com.networknt.schema.JsonSchema.combineCurrentUriWithIds(JsonSchema.java:87) ~[json-schema-validator-1.0.29.jar:?]
```

Note that none of my JSON are referencing one another, nor referencing any external schema through an URN (or any URI for that matter) either. So it is an issue only on the "id" level of JSON schemas being loaded.

I couldn't find out how to make it work with a specific JsonSchemaFactory configuration either?

[...]

PR #906, opened 2023-12-07
665: Can't load JSON schemas with URN value in id field

martin-sladecek:
(Hopefully) a fix for #665.
Added a special handling for "urn:" URIs
When an the same URN is present in some id/anchor, a reference is created. Otherwise, it falls back to the current logic and a special `URIFetcher` needs to be present.


stevehu:
The new release, 1.0.88, is out. Thanks.

PR #910, opened 2023-12-14
Refactor validation message generation

### LINKED ISSUE #744: Race condition using schema configuration when caching in multi-threaded scenarios
Class `SchemaValidatorsConfig` is currently used to define all configuration properties for validation before a `JsonSchema` instance is loaded. This is set on the schema:
* When the schema is initialised.
* If schema caching is enabled, once the schema has been retrieved from the cache.

In the second case, the configuration is set on the schema in case a different configuration was used when the schema was first initialised. This works fine if, using caching, we execute multiple validations sequentially, but is not correct in multi-threaded environments where we have parallel validations (for example a web application). The problem comes from setting the configuration on the cached schema itself which means that it may be changed by one thread while another thread is executing a validation.

As an example of the problem consider a web application using schema caching that allows users to choose the language for their JSON validation report between EN and FR. The following steps take place:
1. Bob selects EN and triggers a validation. The schema is loaded for the first time with EN as its language and cached.
2. Bob's validation begins and starts collecting EN validation messages.
3. Before Bob's validation completes, Alice triggers a validation in FR. The same schema is loaded from the cache and FR is set as the language of choice.
4. Bob's validation continues to produce messages, but given step 3, these are now in FR.

In brief, the issue is one of a shared concurrent cache with mutable state leading to race conditions. The current workaround for this is to either create a new `JsonSchemaFactory` per validation, or use a shared one with caching disabled.

To correct the problem and fully support caching we could follow one of two approaches:
* Option A: When a `JsonSchema` instance is loaded from the cache, clone it in-depth before returning it, and set the configuration on the clone.
* Option B: Use a `ThreadLocal` to set and read the `SchemaValidatorsConfig` instance.

From the two options above, using a `ThreadLocal` would be the simplest and is the approach used also elsewhere for similar concerns (see `CollectorContext`).

### LINKED ISSUE #803: Not able to provide separate custom messages for each property under required or additionalProperties keyword
Currently custom error message support is there per keyword, that means schema can declare/override the default error message for a given keyword's failure under its declaration scope. But for few object level keywords where multiple properties are involved like 'required' or 'additionalProperties', framing the custom message cannot be done at granular level.
Ex: In below scenario, type keyword message can be easily overridden, but for required keyword the custom message cannot be property specific.

```
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "number"
    }
    "bar": {
      "type": "string"
    }
  },
  "required": ["foo", "bar"],
  "additionalProperties": false
  "message": {
    "type" : "Custom Message: Invalid type provided",
    "required": "'foo' and 'bar' both are required"
  }
}
```

Need a way to override error messages for such keywords at property level, like how its supported from ajv - https://ajv.js.org/packages/ajv-errors.html#messages-for-keywords

```
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "number"
    }
    "bar": {
      "type": "string"
    }
  },
  "required": ["foo", "bar"],
  "additionalProperties": false
  "message": {
    "type" : "Custom Message: Invalid type provided",
    "required": {
      "foo" : "'foo' is required",
      "bar" : "'bar' is required"
    }
  }
}
```

### LINKED ISSUE #839: Not able to get field path when I use custom messages for array type.
For other types I can add path manually to custom message, but for arrays I need its index to be printed as well.
Schema Example:
```
"requestedItems": {
  "type": "array",
  "minItems": 1,
  "items": {
    "properties": {
      "item": {
        "type": "string",
        "minLength": 1,
        "message": {
          "minLength": "Item should not be empty"
        }
      }
    }
  }
}
```

As per the above code when I do not put any value in the item field in the request I get below custom message as specified 
**"Item should not be empty"**

But I want custom message to be printed along  with its path, for ex: 
**"$.requestedItems[0].item: Item should not be empty"**

Path gets printed if I remove custom message. A dynamic message created by library would be printed along with field path as shown below. 
**"$.requestedItems[0].item: must be at least 1 characters long"**

But I want the path with custom message. I can add the path along with custom message but for array I can not pass its index. Please help

### LINKED ISSUE #903: Custom error message for missing field
Hello,

I discover recently the way to customize error message using this feature : https://github.com/networknt/json-schema-validator/pull/440

Nevertheless, I don't find a way to define error message in case of missing field in a json message 
For example, if a json schema expect to have an id but the json sent by a user does not contain this id, I want to have a specific error message saying something like "JSON schema not respected, missing id field".
How can this be done ?

Regards,

Tom

### LINKED ISSUE #908: Support list of Locales
It would be great to support multiple locales that will match the first available in the order of preference.
This might be e.g. coming from a client through "Accept-Language" header.

justin-tay:
Resolves #908, resolves #903, resolves #839, resolves #803, resolves #744

This makes the following changes
* Add `MessageSource` interface for deriving localised messages
* Add `ExecutionConfig` to store `Locale` and use it for per-execution localised messages
* Add `Locales` to determine the supported language tags for the default resource bundle and for determining the best matching `Locale`
* Support custom messages for keywords with different properties
* Use `MessageFormat` for formatting custom messages with arguments

### DIFF
--- a/src/main/java/com/networknt/schema/i18n/Locales.java
+++ b/src/main/java/com/networknt/schema/i18n/Locales.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.networknt.schema.i18n;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Locale;
+import java.util.Locale.FilteringMode;
+import java.util.Locale.LanguageRange;
+import java.util.stream.Collectors;
+
+/**
+ * Functions for working with Locales.
+ */
+public class Locales {
+    /**
+     * The list of locale resource bundles.
+     */
+    public static final String[] SUPPORTED_LANGUAGE_TAGS = new String[] { "ar-EG", "cs-CZ", "da-DK", "de", "fa-IR",
+            "fi-FI", "fr-CA", "fr", "he-IL", "hr-HR", "hu-HU", "it", "ja-JP", "ko-KR", "nb-NO", "nl-NL", "pl-PL",
+            "pt-BR", "ro-RO", "ru-RU", "sk-SK", "sv-SE", "th-TH", "tr-TR", "uk-UA", "vi-VN", "zh-CN", "zh-TW" };
+
+    /**
+     * The supported locales.
+     */
+    public static final List<Locale> SUPPORTED_LOCALES = of(SUPPORTED_LANGUAGE_TAGS);
+
+    /**
+     * Gets the supported locales.
+     * 
+     * @return the supported locales
+     */
+    public static List<Locale> getSupportedLocales() {
+        return SUPPORTED_LOCALES;
+    }
+
+    /**
+     * Gets a list of {@link Locale} by language tags.
+     * 
+     * @param languageTags for the locales
+     * @return the locales
+     */
+    public static List<Locale> of(String... languageTags) {
+        return Arrays.asList(languageTags).stream().map(Locale::forLanguageTag).collect(Collectors.toList());
+    }
+
+    /**
+     * Determine the best matching {@link Locale} with respect to the priority list.
+     * 
+     * @param priorityList the language tag priority list
+     * @return the best matching locale
+     */
+    public static Locale findSupported(String priorityList) {
+        return findSupported(priorityList, getSupportedLocales());
+    }
+
+    /**
+     * Determine the best matching {@link Locale} with respect to the priority list.
+     * 
+     * @param priorityList the language tag priority list
+     * @param locales      the supported locales
+     * @return the best matching locale
+     */
+    public static Locale findSupported(String priorityList, Collection<Locale> locales) {
+        return findSupported(LanguageRange.parse(priorityList), locales, FilteringMode.AUTOSELECT_FILTERING);
+    }
+
+    /**
+     * Determine the best matching {@link Locale} with respect to the priority list.
+     * 
+     * @param priorityList  the language tag priority list
+     * @param locales       the supported locales
+     * @param filteringMode the filtering mode
+     * @return the best matching locale
+     */
+    public static Locale findSupported(List<LanguageRange> priorityList, Collection<Locale> locales,
+            FilteringMode filteringMode) {
+        Locale result = Locale.lookup(priorityList, locales);
+        if (result != null) {
+            return result;
+        }
+        List<Locale> matching = Locale.filter(priorityList, locales, filteringMode);
+        if (!matching.isEmpty()) {
+            return matching.get(0);
+        }
+        return Locale.ROOT;
+    }
+}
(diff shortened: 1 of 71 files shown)


src/main/java/com/networknt/schema/i18n/Locales.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Locale.FilteringMode;
+import java.util.Locale.LanguageRange;
+import java.util.stream.Collectors;
+
+/**
+ * Functions for working with Locales.
+ */
+public class Locales {
```
martin-sladecek, line 29: If I read it correctly, this class doesn't work that well with custom `MessageSource`s and you need to know what bundle is configured and use the correct list of supported Locales.

What about integrating this functionality to the `MessageSource` instead? The source should know what languages it can support and it would be responsible for choosing the `Locale`.
Then in the `ExecutionContext`/``ExecutionConfig`, you'd set only the priority list and it would get resolved through the `MessageSource` automatically.

justin-tay: I prefer to separate the concerns since the use of a priority list is really an optional thing. 

Attempting to implement it in the `MessageSource` means that that it needs to keep evaluating the priority list each time for each message which won't be efficient.

If you have a custom `MessageSource` then you would know what locales you support. Then you can just use `Locales.findSupported(priorityList, mySupportedLocalesForMyMessageSource)` and just use that Locale to set in the `ExecutionConfig`.

Also you might come from a framework that already implements some of these parts, for instance Spring and then you can just use those instead, for instance Spring has a [LocaleResolver](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/LocaleResolver.html) and [MessageSource](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/MessageSource.html).

martin-sladecek: OK, makes sense to leave the locale resolution on the client. The resolution has to be done in a bigger context (let's say of the whole app), so that the translation is consistent. 

But I still think having the list of supported languages would be better nicely coupled inside the `MessageSource`.
I think it's not ideal to rely on the fact that the developer knows what is being used where, it's better if there's a possibility to can get the information through a single API when needed. And it shouldn't matter if you use the default bundle or a custom bundle, there should be one API for one task. 
 
E.g. if this library becomes a part of something more complex, I can imaging some piece of a (utility) code that prepares some ready-made `JsonSchema` with some custom bundle. And a different piece of code for interoperability between this library and some other framework (like Spring classes that you mentioned), let's say this code is there to choose the correct locale in you app and automatically set the best candidate for your json schema validation.
These pieces can be coming from different libraries/developers and ideally, they should work together.

justin-tay: The following is the typical usage

```java
SchemaValidatorsConfig config = new SchemaValidatorsConfig();
JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V201909);
JsonSchema schema = factory.getSchema(source, config);

String accept = "it;q=0.9,fr;q=1.0";
Locale locale = Locales.findSupported(accept);
ExecutionContext executionContext = jsonSchema.createExecutionContext();
executionContext.getExecutionConfig().setLocale(locale);
Set<ValidationMessage> messages = jsonSchema.validate(executionContext, rootNode);
```

You can always build your own abstraction on top of this if you want to prepare a read-made `JsonSchema` with some specific configuration and some custom bundle to hide the implementation details. This will also let you switch out to a different json schema validator if need be.

martin-sladecek: Yes, I understand that the usage, I've seen it in the new documentation, I was just suggesting to put the list of locales to the MessageSource, because it can be considered a property of a message source. Of course, one can make abstractions without that, but IMO it would be more convenient.

stevehu: @justin-tay Just want to confirm with you if this change is a breaking change. If yes, we need to make sure that we increase at least a minor number for the next release. Thanks.

stevehu (approved):

src/main/java/com/networknt/schema/i18n/Locales.java
justin-tay: Yes, this and in particular the changes in subsequent pull requests I have made have breaking changes. Thanks for your effort in maintaining this project, it's much appreciated!

PR #912, opened 2023-12-14
Downgrade logback to 1.3.14 for JDK 8 compatibility

### LINKED ISSUE #911: CI failing as Logback 1.4 requires min JDK 11 but the project still retains JDK 8 compatibility
Logback 1.3 which is [compatible with JDK 8](https://logback.qos.ch/news.html) should probably be used instead if JDK 8 compatibility is required.

Related issue
- https://github.com/networknt/json-schema-validator/issues/907

justin-tay:
Resolves #911


stevehu:
@justin-tay Thanks a lot for your help.

PR #913, opened 2023-12-14
Update documentation on CollectorContext

justin-tay:
Updates the documentation to reflect that the `CollectorContext` is no longer stored in a `ThreadLocal` and that the `CollectorContext` needs to be explicitly reused to collect across multiple runs.


stevehu:
@justin-tay Thanks a lot for your help.

PR #934, opened 2024-01-26
fixes #933 update javadoc and a test case

### LINKED ISSUE #933: update javadoc and a test case

stevehu:


stevehu:
@justin-tay, One of the test cases failed on my faster desktop computer during the maven install and I am guessing it is a timing issue with other test cases. I have added ten milliseconds of sleep as a temporary solution. Do you happen to know what would cause IlegalStateException "recursive update"? If not, I think we can leave the test case as it is. Thanks a lot for the PR.

justin-tay (approved):
Looks good to me. Thanks!

I shall see if I can reproduce this later.

PR #946, opened 2024-02-02
Add test for type loose for array and update doc for behavior

### LINKED ISSUE #415: When typeLoose set to true, validation convert integers/string to array
According to documentation typeLoose is used to allow sending string as integers/numbers ("When typeLoose is true, the validator will convert strings to different types to match the type defined in the schema"), but in fact it effects other types validation - not only number/integers.
In my example the schema has array of integers and array of objects:
```
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "http://example.com/example.json",
  "type": "object",
  "properties": {
    "array_of_integers": {
      "$id": "#/properties/array_of_integers",
      "type": "array",
      "items": {
        "type": "integer"
      }
    },
    "array_of_objects": {
      "$id": "#/properties/array_of_objects",
      "type": "array",
      "items": {
        "type": "object"
      }
    }
  }
}
```

The following json pass validation (when typeLoose = true):
```
{
  "array_of_integers": 1,
  "array_of_objects": {}
}
```

I expect it to fail (like when typeLoose = false):
```
$.array_of_integers: integer found, array expected
$.array_of_objects: object found, array expected
```

justin-tay:
Closes #415

Add test for type validator type loose behavior for arrays and update documentation accordingly that type loose will interpret an item as an array of size 1 and that it will then only validate that that item matches the type defined for the array.

Not sure if there is supposed to be a spec which defines what type loose behavior is supposed to be.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -22,7 +22,7 @@ JsonSchema jsonSchema = JsonSchemaFactory.getInstance().getSchema(schema, config
 
 * typeLoose
 
-When typeLoose is true, the validator will convert strings to different types to match the type defined in the schema. This is mostly used to validate the JSON request or response for headers, query parameters, path parameters, and cookies. For the HTTP protocol, these are all strings and might be defined as other types in the schema. For example, the page number might be an integer in the schema but passed as a query parameter in string. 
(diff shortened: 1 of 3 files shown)


stevehu:
@justin-tay Type-loose is not a concept in the specification but an extra feature in the library. This was introduced to validate incoming requests against the OpenAPI specifications. In HTTP requests, headers, query and path parameters don't have type information.

Issue #475, opened 2021-10-26
Error validating a JSON Schema against formal draft JSON Schema meta data


ericbroda:
My objective is to validate a JSON Schema against JSON Schema meta-data (ie. use the JSON schema meta-data as the "grammar" to validate an actual JSON Schema)

The issue is that obvious errors are not caught:
- changing the type "array" to "sdfsdfsd" - no error caught
- changing a "$ref" to point to a non-existing definition - no error caught

I have provided several files below:
- jsonschema.json: this is the combined meta-data files from json-schema.org
- test.schema.json: this is the schema to be validated (against jsonschema.json)
- cde fragment to execute the validation.

(In the code fragment (later), this is the file "test.schema.json")
~~~~
{
  "$id": "https://example.com/arrays.schema.json",
  "description": "A representation of a person, company, organization, or place",
  "type": "object",
  "required": [ "messages" ],
  "properties": {
    "messages": {
      "type": "array",
      "items": { "$ref": "#/$defs/message" }
    }
  },
  "$defs": {
    "message": {
      "type": "object",
      "required": [ "key", "data" ],
      "properties": {
        "key": {
          "$ref": "#/$defs/key"
        },
        "data": {
          "$ref": "#/$defs/data"
        }
      }
    },
    "key": {
      "type": "string"
    },
    "data": {
      "type": "object",
      "required": [ "veggieName", "veggieLike" ],
      "properties": {
        "veggieName": {
          "type": "string",
          "description": "The name of the vegetable."
        },
        "veggieLike": {
          "type": "boolean",
          "description": "Do I like this vegetable?"
        }
      }
    }
  }
}
~~~~

The JSON Schema meta-data is a consolidation of meta-data files from JSON-Schema.org (https://json-schema.org/specification.html):

(in the code below, this is "jsonschema.json)
~~~~
{
    "$id": "https://json-schema.org/draft/2020-12/meta/validation",
    "$vocabulary": {
        "https://json-schema.org/draft/2020-12/vocab/validation": true
    },
    "$dynamicAnchor": "meta",

    "title": "Validation vocabulary meta-schema",
    "type": ["object", "boolean"],
    "properties": {
        // validation.json
        "type": {
            "anyOf": [
                { "$ref": "#/$defs/simpleTypes" },
                {
                    "type": "array",
                    "items": { "$ref": "#/$defs/simpleTypes" },
                    "minItems": 1,
                    "uniqueItems": true
                }
            ]
        },
        "const": true,
[...]

PR #980, opened 2024-03-01
Support custom vocabularies and unknown keyword and meta-schema handling

### LINKED ISSUE #829: Schema version no longer validated since version 1.0.83
I noticed that sice the release of version 1.0.83 the behaviour changed, and wrong schema versions do no longer throw an exception.

With version 1.0.82 the following Kotlin example code would have thrown a `JsonSchemaException` because the schema version doesn't match:
```kotlin
val schemaNode = jacksonObjectMapper().readTree("{\"\$schema\": \"http://json-schema.org/draft-06/schema\", \"properties\": { \"id\": {\"type\": \"number\"}}}")
val factory = com.networknt.schema.JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V202012)
val schema = factory.getSchema(schemaNode)
```

Now, since version 1.0.83 this is no longer the case and regardless what version flag is specified, all schema versions are accepted.

### LINKED ISSUE #977: Are there plans to support custom vocabularies?
Currently, the vocabulary set is limited to https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/Vocabulary.java#L30

Are there plans to support custom vocabularies that consumers can define?

justin-tay:
Closes #977, closes #829

This adds support for customizing
* Custom vocabularies
* Unknown keyword handling
* Unknown meta-schema handling

This contains breaking changes in how custom meta-schemas are created  and the documentation updates assume it will be released as 1.4.0.

`JsonSchemaFactory`
* The following were renamed on `JsonSchemaFactory` builder
  * `defaultMetaSchemaURI` -> `defaultMetaSchemaIri`
  * `enableUriSchemaCache` -> `enableSchemaCache`
* The builder now accepts a `JsonMetaSchemaFactory` which can be used to restrict the loading of meta-schemas that aren't explicitly defined in the `JsonSchemaFactory`. The `DisallowUnknownJsonMetaSchemaFactory` can be used to only allow explicitly configured meta-schemas.

`JsonMetaSchema`
* In particular `Version201909` and `Version202012` had most of the keywords moved to their respective vocabularies.
* The following were renamed
  * `getUri` -> `getIri`
* The builder now accepts a `vocabularyFactory` to allow for custom vocabularies.
* The builder now accepts a `unknownKeywordFactory`. By default this uses the `UnknownKeywordFactory` implementation that logs a warning and returns a `AnnotationKeyword`. The `DisallowUnknownKeywordFactory` can be used to disallow the use of unknown keywords.
* The implementation of the builder now correctly throws an exception for `$vocabulary` with value of `true` that are not known to the implementation.

`ValidatorTypeCode`
* `getNonFormatKeywords` has been removed and replaced with `getKeywords`. This now includes the `format` keyword as the `JsonMetaSchema.Builder` now needs to know if the `format` keyword was configured, as it might not be in meta-schemas that don't define the format vocabulary.
* The applicable `VersionCode` for each of the `ValidatorTypeCode` were modified to remove the keywords that are defined in vocabularies for `Version201909` and `Version202012`.

`Vocabulary`
* This now contains `Keyword` instances instead of the string keyword value as it needs to know the explicit implementation. For instance the implementation for the `items` keyword in Draft 2019-09 and Draft 2020-12 are different.
* The following were renamed
  * `getId` -> `getIri`


rishabh413:
Hi @justin-tay 

Will this also fix https://github.com/networknt/json-schema-validator/issues/475#issuecomment-1618381541 ?

justin-tay:
> Hi @justin-tay
> 
> Will this also fix [#475 (comment)](https://github.com/networknt/json-schema-validator/issues/475#issuecomment-1618381541) ?

That should already have been fixed as it has nothing to do with implementing custom vocabularies. That issue has to do with the lack of support for `$dynamicRef` which was fixed in

- #931

rishabh413:
It did, thanks!!

PR #1002, opened 2024-03-28
Resolved smells.

meer1616:
Fix the implementation smells in ExclusiveMaximumValidator.java by reducing complex condition and segregating them into different methods, added another method called addFieldsToMappingWithCheck in BaseJsonValidator class, which adds the field to mapping by doing some null checking which reduces the cyclomatic complexity of registerAndMergeDiscriminator method, and make it clear and understandable and Lastly changed the method name from "from" to "extractSchemaRefFromSchema" in JsonSchemaRef which makes it more descriptive. The method name explains what it does and helps the new contributor understand the method's purpose without going through the whole code.

### DIFF
--- a/src/main/java/com/networknt/schema/BaseJsonValidator.java
+++ b/src/main/java/com/networknt/schema/BaseJsonValidator.java
@@ -71,7 +71,8 @@ public abstract class BaseJsonValidator extends ValidationMessageHandler impleme
         this.validationContext = copy.validationContext;
     }
 
-    private static JsonSchema obtainSubSchemaNode(final JsonNode schemaNode, final ValidationContext validationContext) {
+    private static JsonSchema obtainSubSchemaNode(final JsonNode schemaNode,
+            final ValidationContext validationContext) {
         final JsonNode node = schemaNode.get("id");
 
         if (node == null) {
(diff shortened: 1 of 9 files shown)


justin-tay:
I actually don't really see the need to rename `from` to `extractSchemaRefFromSchema`. The naming convention of using `from` or `of` is quite common. eg. https://docs.oracle.com/javase/8/docs/api/java/time/Period.html#from-java.time.temporal.TemporalAmount-.

The changes from `if ((schemaNode.isLong() || schemaNode.isInt()) && (JsonType.INTEGER.toString().equals(getNodeFieldType()))) {` to `if (checkSchemaNodeLongOrInt(schemaNode) && checkIntEqNodeFieldType()) {` also seem rather arbitrary and I don't really see this adding any readability or promote reuse.

This only leaves the refactor of `registerAndMergeDiscriminator` which seems ok, but again rather arbitrary to pull the method out as the method is private and thus doesn't add any benefit of testability.

src/main/java/com/networknt/schema/utils/JsonSchemaRefs.java
```diff
@@ -33,7 +33,7 @@ public class JsonSchemaRefs {
      * @param schema the schema
      * @return the ref
      */
-    public static JsonSchemaRef from(JsonSchema schema) {
+    public static JsonSchemaRef extractSchemaRefFromSchema(JsonSchema schema) {
```
justin-tay, line 36: I actually don't really see the need to rename `from` to `extractSchemaRefFromSchema`. The naming convention of using `from` or `of` is quite common. eg. https://docs.oracle.com/javase/8/docs/api/java/time/Period.html#from-java.time.temporal.TemporalAmount-.

src/main/java/com/networknt/schema/ExclusiveMaximumValidator.java
```diff
@@ -35,13 +35,22 @@ public class ExclusiveMaximumValidator extends BaseJsonValidator {
...
+
+
     public ExclusiveMaximumValidator(SchemaLocation schemaLocation, JsonNodePath evaluationPath, final JsonNode schemaNode, JsonSchema parentSchema, ValidationContext validationContext) {
         super(schemaLocation, evaluationPath, schemaNode, parentSchema, ValidatorTypeCode.EXCLUSIVE_MAXIMUM, validationContext);
         if (!schemaNode.isNumber()) {
             throw new JsonSchemaException("exclusiveMaximum value is not a number");
         }
         final String maximumText = schemaNode.asText();
-        if ((schemaNode.isLong() || schemaNode.isInt()) && (JsonType.INTEGER.toString().equals(getNodeFieldType()))) {
+        if (checkSchemaNodeLongOrInt(schemaNode) && checkIntEqNodeFieldType()) {
```
justin-tay, line 53: The changes from `if ((schemaNode.isLong() || schemaNode.isInt()) && (JsonType.INTEGER.toString().equals(getNodeFieldType()))) {` to `if (checkSchemaNodeLongOrInt(schemaNode) && checkIntEqNodeFieldType()) {` also seem rather arbitrary and I don't really see this adding any readability or promote reuse.

src/main/java/com/networknt/schema/BaseJsonValidator.java
```diff
@@ -137,67 +142,83 @@ && noExplicitDiscriminatorKeyOverride(discriminatorMapping, jsonSchema)) {
...
-
-                    final JsonNode currentMappingValue = mappingOnContextDiscriminator.get(mappingKeyToAdd);
-                    if (null != currentMappingValue && currentMappingValue != mappingValueToAdd) {
-                        throw new JsonSchemaException(instanceLocation + "discriminator mapping redefinition from " + mappingKeyToAdd
-                                + "/" + currentMappingValue + " to " + mappingValueToAdd);
-                    } else if (null == currentMappingValue) {
-                        mappingOnContextDiscriminator.set(mappingKeyToAdd, mappingValueToAdd);
-                    }
-                }
+                addFieldsToMappingWithCheck(fieldsToAdd, mappingOnContextDiscriminator, instanceLocation);
```
justin-tay, line 202: This refactor seems rather arbitrary to pull the method out, as the method is private and thus doesn't add any benefit of testability.

PR #1006, opened 2024-03-30
refactored copySchemaResources with polymorphism

meer1616:
The conditional logic in `copySchemaResources` method has been replaced with polymorphic behavior. Each type of resource copying is now handled by a separate subclass of ResourceCopier, promoting cleaner code and easier maintenance.


justin-tay:
This is not how polymorphism works and this increases garbage collection for no benefit.

PR #1007, opened 2024-03-30
Default class and test cases.

meer1616:
Changed the class name to Default and added the test cases for the same.


justin-tay:
Why are you opening a new pull request?

Duplicate of #1005

PR #1018, opened 2024-04-16
Add options to control caching of schemas

### LINKED ISSUE #1016: Recursive calls leading to an OOM when parsing a valid schema
Hello!

I've been trying to upgrade our project to the latest version but detected an issue where the validator causes an OOM with a schema that previously worked just fine (and validates against other on-line validators). This commit in my fork reproduces the problem:

https://github.com/dweiss/json-schema-validator/commit/cefbf5bb4a6c2310d7c60c1965c4a4dcfc417d7e

It seems to be somewhere around the loop initializing validators:
image

It's not clear to my why this doesn't end with a stack overflow - instead, it just leads to an OOM, almost as if it were fanning out somewhere (there are lots of refs in that schema).

I didn't dig deep but I thought you folks would be interested and would have more expertise to tell what's going on. Thank you.

justin-tay:
Closes #1016

This adds additional configuration options to control the caching / preloading schema behavior.

* Option to not cache schema `$ref`. This means that after resolving a `$ref` during evaluation the schema loaded is discarded and won't be used for subsequent evaluation runs. This is set to a default of `true` to cache by default. This may be needed if there is a large permutation of evaluation paths due to multiple nested applicators like `allOf`, `oneOf` and `anyOf` which makes caching all evaluation paths take up too much memory.
* Option to limit the max nesting depth when preloading `$ref`. This is set to a reasonable default of `40`.


stevehu (approved):

txshtkckr:
Any chance we could get a release with this option in it? I'm encountering this issue in production use.

On a related note, looking at `CachedSupplier`, it holds on to its `delegate` indefinitely. I have not dug into the library enough to be sure if there are thread-safety concerns around making both `delegate` mutable and `null`-ing it out after calling it, but that would allow it to release all the references captured in the lambda at the call point, which would greatly reduce the memory pressure that this causing. In a heap dump that I analyzed, the retained set from the `RefValidator` was 450M, and even a very sprawled eval path tree shouldn't have this kind of impact. But since the supplier captures the validation context, the heap I'm looking at shows it having captured 90,000 of those, and the impact of **that** is definitely not trivial.

txshtkckr:
If thread-safety **is** an issue, an easy workaround would be to change the supplier field to `volatile` and write back a constant supplier after performing the calculation:

```
    @Override
    public T get() {
        if (cache == null) {
            T value = delegate.get();
            cache = value;
            delegate = () -> value;
        }
        return cache;
    }
```

This would release whatever garbage the supplier is holding while still providing the guarantee that the delegate is never `null`.

PR #1021, opened 2024-04-16
i18 support for spanish and english locales are added

channaveer1:
these two are needed for us and we are using it in our production please merge asap and provide the release


src/main/resources/jsv-messages_en.properties
```diff
@@ -0,0 +1,70 @@
+$ref = {0}: has an error with ''refs''
```
justin-tay, line 1: The 0 byte `jsv-messages_en.properties` should already fallback on `jsv-messages.properties`. Why is this needed?

justin-tay (changes requested):

src/main/resources/jsv-messages_es.properties
```diff
@@ -0,0 +1,70 @@
+$ref = {0}: tiene un error con ''refs''
```
justin-tay, line 1: The `encoding` test in `LocaleTest` needs to be modified to ensure that the encoding of this file works for Java 8.

justin-tay (changes requested):

src/main/resources/jsv-messages_en.properties
channaveer1: Okay thats right

src/main/resources/jsv-messages_es.properties
channaveer1: this is taken care

channaveer1: Could you please le me know how to do that?

justin-tay: You need to modify this test.

https://github.com/networknt/json-schema-validator/blob/ffec188f5d0b5d4f1190a56f8cbfb48b4c15a89d/src/test/java/com/networknt/schema/LocaleTest.java#L118-L125

You also need to modify this to add the supported locale

https://github.com/networknt/json-schema-validator/blob/ffec188f5d0b5d4f1190a56f8cbfb48b4c15a89d/src/main/java/com/networknt/schema/i18n/Locales.java#L33-L35

src/test/java/com/networknt/schema/LocaleTest.java
```diff
@@ -145,11 +145,13 @@ void encoding() {
         expected.put("vi","$: phải dài tối đa 5 ký tự");
         expected.put("zh_CN","$: 长度不得超过 5 个字符");
         expected.put("zh_TW","$: 長度不得超過 5 個字元");
+        expected.put("es","$: debe tener como máximo 5 caracteres");
 
         // In later JDK versions the numbers will be formatted
         Map<String, String> expectedAlternate = new HashMap<>();
         expectedAlternate.put("ar","$: يجب أن يكون طوله ٥ حرفًا على الأكثر");
         expectedAlternate.put("fa","$: باید حداکثر ۵ کاراکتر باشد");
+        expectedAlternate.put("es","$: debe tener como mÃ¡ximo 5 caracteres");
```
justin-tay, line 154: What is this comparison for and how is this possible? The rest of the alternate comparisons is because later JDK versions perform mapping to arabic numerals.

justin-tay (changes requested):

src/test/java/com/networknt/schema/LocaleTest.java
justin-tay: Note that for Java 8 the properties file must be encoded in ISO 8859-1. UTF-8 is only supported from Java 9 onwards. If the test fails for Java 8 but passes for the rest it means that your file is not encoded in ISO 8859-1.

channaveer1: @justin-tay 
in the English text "Hello, how are you?" is encoded to ISO 8859-1 and then decoded back to a string. However, since ISO 8859-1 does not include characters specific to Spanish, the output will be the same as the input.

justin-tay: If the encoding is correct then remove the expectedAlternate line and the test should pass. If not it's just not encoded correctly and is being corrupted to this `mÃ¡ximo` text. This means that anyone using Java 8 will see `debe tener como mÃ¡ximo 5 caracteres` as the translated text which is obviously not correct.

channaveer1: @justin-tay do you prefer any tool for conversion. so that i can verify once more.

justin-tay: I don't prefer any tool, it just need to be encoded such that Java 8 can load it properly. I just use Eclipse which will automatically convert text into ISO 8859-1 when pasted into a properties file as that is the default configuration in Eclipse for the Java Properties File content type.

channaveer1: @justin-tay i tried the same i set my intelij editor to ISO 8859-1 i commited  now still java 8 is failing. can you please do me a favour for this

channaveer1: @justin-tay can you please give me the converted file. its needed for us

channaveer1: @justin-tay please let me any help.

PR #1043, opened 2024-05-28
Update example in javadoc for JsonMetaSchema builder

khouari1:


stevehu (approved):

justin-tay:
The fix actually isn't quite right either. It should be `JsonMetaSchema.getV4()` and not `JsonMetaSchema.getDraftV4()`.

Issue #185, opened 2019-09-09
Validation issue in oneOf when elements have optional fields


ddobrin:
Currently, oneOf validates each element in the schema.

2 problems need to be addressed:
- prevent more than one element to be passed in oneOf - violation of OpenAPI spec
- address the PropertiesValidator, which can not distinguish whether a missing node is an optional field or one of the oneOf elements which is not present in the request

Issue #110, opened 2018-12-19
Validation Error when using OneOf in OpenAPI specs


ddobrin:
The OneOf Validator iterates through the validators for each choice in OneOf and validates.
When it encounters the second "valid" schema element, it returns a validation error:

{ "error" : {"statusCode":400,"code":"ERR11004","message":"VALIDATOR_SCHEMA","description":"Schema Validation Error - $.search: should be valid to one and only one of the schemas ","severity":"ERROR"} }

In this case, the matching element was the 8th out of 11th, however the code analyzed them one by one and failed in the second one already, where only the 8th is a true match.

The issue is in the PropertiesValidator, where it tries to find an entry by a key and does not find it.
It treats the "not found" as normal, instead of issuing an error and letting the OneOf Validator handle it gracefully (in this ignore it)
...
            JsonNode propertyNode = node.get(entry.getKey());

            if (propertyNode != null) {
                errors.addAll(propertySchema.validate(propertyNode, rootNode, at + "." + entry.getKey()));
...

Proposed resolution includes the issuance of an error when property node does not match and letting the caller in the recursive chain handle it.

It must be properly validated and ascertained that there are no side-effects.

It is a showstopper for a number of APIs at a client site.

ddobrin:
![](https://github.trello.services/images/mini-trello-icon.png) [json-schema-validator: fixed #110 Validation Error when using OneOf in OpenAPI specs](https://trello.com/c/a0OqlzAW/50-json-schema-validator-fixed-110-validation-error-when-using-oneof-in-openapi-specs)

stevehu:
From the OpenAPI 3.0 specification document, oneOf – validates the value against exactly one of the subschemas. If you have two or more subschemas valid, then an error should be thrown.

PR #1053, opened 2024-06-09
Remove incorrect logic for oneOf, anyOf and properties

justin-tay:
This removes the incorrect logic introduced as part of the fixes for
* #110
* #185 

This incorrect logic applies to applicators such as `oneOf` such that it will attempt to special handle cases where there is no validation failures but there are optional properties that are missing.

There is actually no special logic differences between JSON Schema and OpenAPI with regards to the handling of applicators such as `oneOf`. This misunderstanding likely arose due to the incorrect example for `oneOf` on one of the swagger websites which has not been fixed. The OpenAPI spec authors has clarified that the example is wrong and is not compliant with the OpenAPI spec.

* https://github.com/swagger-api/swagger.io/issues/253
* https://github.com/OAI/OpenAPI-Specification/issues/3477


s-ven:
Hum, should this change have been part of a MINOR revision change ?

Every other release, our validation workflow breaks and we have to refactor our schemas to upgrade to a more recent MINOR revision of networknt

stevehu:
I didn't realize that the changes broke the user's code. What is the best way to have the next release? Should we remove the 1.4.1, 1.4.2 and 1.4.3 and release 1.5.0?

justin-tay:
I think it's too late to remove 1.4.1, 1.4.2, 1.4.3 as now that will break the libraries/users that are already using those versions.

I actually consider this change to be a bug fix. It should be noted that none of the existing tests nor even the additional tests actually hit the affected removed codes. Even the invalid swagger example actually behaved like the standard applicator, ie. the swagger example was already no longer applicable and didn't consider nodes that failed with required to be acceptable, likely this was fixed by other changes as these codes aren't specific to openapi and would have affected the standard json schema evaluations. This possibly introduced very subtle hard to find bugs in standard evaluations.

s-ven:
agree, too late to remove, that's not what I'm asking,
I just wanted you guys to know about this (it happened also when I migrated from 1.3.0 to 1.3.x, I had to refactor the schema builder code)
thanks for prompt answer, and also for building this library !

stevehu:
@s-ven I have merged all the latest PRs and released the 1.5.0 tag. Thanks a lot for the suggestions. It is hard to determine if any change can break users' code as different users will use the library differently. It is important for our users to report these breaking changes to us. Thanks a lot for your help.

PR #1056, opened 2024-06-11
Release delegate held by cached supplier

justin-tay:
Change to allow the delegate to be garbage collected.


txshtkckr:
Although allowing the cached references to be disabled entirely is an even bigger memory win, for completeness I'll offer the results of a simple test that I ran using the JSON schema we use for the Atlassian Document Format, which is available at https://go.atlassian.com/adf-json-schema .

* Loading the schema and validating a simple document with version 1.4.3: ~455MB retained set
* The same using version 1.5.0: ~150MB
* The same using version 1.5.0 with `cachedRefs(false)`: ~3MB

While we will still need to disable the caching in our own case, releasing the delegate was definitely worth it. Thanks!

PR #1068, opened 2024-06-19
Add builder for `SchemaValidatorsConfig`

justin-tay:
This creates a `SchemaValidatorsConfig.Builder` to create `SchemaValidatorsConfig` instances. This also updates the documentation and release notes assuming that the next release will be `1.4.1`. This also updates all the tests that used the constructor to the builder.

This is was done in order to
* Eventually make `SchemaValidatorsConfig` immutable. For now this just deprecates the `SchemaValidatorsConfig` constructor as this part of the public API is quite widely used. Instances by the builder will throw `UnsupportedOperationException` when setters are called.
* Make changes to the defaults in `SchemaValidatorsConfig`. For instance setting `PathType.JSON_POINTER` as default as this aligns better with the spec. The changes in defaults only apply if the builder was used to construct the instance so existing codes using the constructor are unaffected and can migrate when it suits them.

Note that there are differences in defaults from the builder vs the constructor.

The following builder creates the same values as the constructor previously.

```java
SchemaValidatorsConfig config = SchemaValidatorsConfig.builder()
    .pathType(PathType.LEGACY)
    .errorMessageKeyword("message")
    .nullableKeywordEnabled(true)
    .build();
```

The following configurations were renamed with the old ones deprecated
* `handleNullableField` -> `nullableKeywordEnabled`
* `openAPI3StyleDiscriminators` -> `discriminatorKeywordEnabled`
* `customMessageSupported` -> `errorMessageKeyword`

The following defaults were changed in the builder vs the constructor
* `pathType` from `PathType.LEGACY` to `PathType.JSON_POINTER`
* `handleNullableField` from `true` to `false`
* `customMessageSupported` from `true` to `false`

When using the builder custom error messages are not enabled by default and must be enabled by specifying the error message keyword to use ie. "message".

| Deprecated Code                                                        | Replacement
|------------------------------------------------------------------------|----------------------------------------------------------------------
| `SchemaValidatorsConfig config = new SchemaValidatorsConfig();`        | `SchemaValidatorsConfig config = SchemaValidatorsConfig().builder().pathType(PathType.LEGACY).errorMessageKeyword("message").nullableKeywordEnabled(true).build();`
| `config.setEcma262Validator(true);`                                    | `builder.regularExpressionFactory(JoniRegularExpressionFactory.getInstance());`
| `config.setHandleNullableField(true);`                                 | `builder.nullableKeywordEnabled(true);`
| `config.setOpenAPI3StyleDiscriminators(true);`                         | `builder.discriminatorKeywordEnabled(true);`
| `config.setCustomMessageSupported(true);`                              | `builder.errorMessageKeyword("message");`


stevehu:
@justin-tay Please let me know if we should release the 1.4.1 tag. Thanks.

justin-tay:
@stevehu Think we should release the 1.4.1 tag. Thanks!

PR #1095, opened 2024-07-25
Fix enum validator messages for object, array and text nodes

### LINKED ISSUE #1094: Missing error text details for values of enumeration when enum type is JSON object
Hello.

I have a JsonSchema with an enumeration of JSON objects:
```json
{
    "title": "Severity",
    "type": "object",
    "properties": {
        "name": {
            "title": "Name",
            "description": "The human readable name of the severity",
            "type": "string"
        },
        "cardinality": {
            "title": "Cardinality",
            "description": "The severities cardinality, the higher the worse it gets",
            "type": "integer",
            "minimum": 0,
            "maximum": 50,
            "multipleOf": 10
        }
    },
    "additionalProperties": false,
    "required": [
        "name",
        "cardinality"
    ],
    "enum": [
        {
            "name": "EMPTY",
            "cardinality": 0
        },
        {
            "name": "OK",
            "cardinality": 20
        },
        {
            "name": "UNKNOWN",
            "cardinality": 30
        },
        {
            "name": "WARNING",
            "cardinality": 40
        },
        {
            "name": "CRITICAL",
            "cardinality": 50
        }
    ],
    "default": {
        "name": "UNKNOWN",
        "cardinality": 30
    }
}
```

When I now validate the following JSON against that:
```json
{
   "name": "FOO",
   "cardinality": 50
}
```

I get a validation error, but with the following error:
```
enum -> does not have a value in the enumeration [, , , , ]
```

Seems like only scalar JSON values are contained in the generated error text, could that be?

justin-tay:
Closes #1094 

This fixes the enum validator messages as previously object and array nodes were empty strings and the text nodes outputted text without the quotes.

For example given the schema

```json
{
  "enum": [6, "foo", [], true, {"foo": 12}]
}
```

The message is now

`: does not have a value in the enumeration [6, "foo", [], true, {"foo":12}]`

### DIFF
--- a/src/main/java/com/networknt/schema/EnumValidator.java
+++ b/src/main/java/com/networknt/schema/EnumValidator.java
@@ -37,6 +37,14 @@ public class EnumValidator extends BaseJsonValidator implements JsonValidator {
     private final Set<JsonNode> nodes;
     private final String error;
 
+    static String asText(JsonNode node) {
+        if (node.isObject() || node.isArray() || node.isTextual()) {
+            // toString for isTextual is so that there are quotes
+            return node.toString();
+        }
+        return node.asText();
+    }
+    
     public EnumValidator(SchemaLocation schemaLocation, JsonNodePath evaluationPath, JsonNode schemaNode, JsonSchema parentSchema, ValidationContext validationContext) {
         super(schemaLocation, evaluationPath, schemaNode, parentSchema, ValidatorTypeCode.ENUM, validationContext);
         if (schemaNode != null && schemaNode.isArray()) {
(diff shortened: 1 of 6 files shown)


jochenberger:
IMHO, this kind of change should not have been made in a patch-level release.

stevehu:
@jochenberger The change is a bug fix and does not break any API. That is why it is a patch release. We can bump up the minor version if there is a reason. Thanks.

stevehu:
As we have added double quotes into the error message, it might break some applications that use the validation message as part of the JSON object. Some escapes are necessary to handle the double quote or back-slash in the validation message.

PR #1124, opened 2024-10-29
Fix hasAdjacentKeywordInEvaluationPath optimisation

### LINKED ISSUE #1123: `unevaluatedProperties` not evaluating properties stemming from `$ref` schemas loaded through nested `allOf`
I believe I found a bug when determining `unevaluatedProperties` (version 1.5.2). For now, I'll show where I run into issues in context of my actual codebase. My hope is this can already convey expectations vs. actual result. I can try creating a small repro later.

Take the follow schema:

```
{
  "$schema": "https://json-schema.org/draft/2019-09/schema",
  "type": "object",
  "allOf": [ { "$ref": "PrimaryDeviceConfiguration.json#PrimaryDeviceConfiguration" } ],
  "properties": {
    "__type": { "const": "dk.cachet.carp.common.application.devices.Smartphone" }
  },
  "unevaluatedProperties": false,
  "$defs": {
    "DeviceRegistration": {
      "$anchor": "DeviceRegistration",
      "$ref": "DefaultDeviceRegistration.json"
    }
  }
}
```

`PrimaryDeviceConfiguration.json#PrimaryDeviceConfiguration` in turn has a nested `allOf`:

```
    "PrimaryDeviceConfiguration": {
      "$anchor": "PrimaryDeviceConfiguration",
      "allOf": [ { "$ref": "DeviceConfiguration.json#DeviceConfiguration" } ],
      "properties": {
        "isPrimaryDevice": { "const": true }
      },
      "required": [ "isPrimaryDevice" ]
    }
  }
```

`DeviceConfiguration.json#DeviceConfiguration` in turn defines a `roleName` property.

The base schema referenced (`PrimaryDeviceConfiguration.json#PrimaryDeviceConfiguration`) doesn't set `unevaluatedProperties` anywhere. The intent is for the base types to allow extension, but _this specific type_, defined in this schema, shouldn't have any additional unknown properties.

However, with this setup, I get warnings on properties which are defined through my `allOf` schema.

> 'roleName' is not evaluated and the schema does not allow unevaluated properties

When I add this property directly to `allOf` in the first schema, the error is gone.

```
  "allOf": [
    { "$ref": "PrimaryDeviceConfiguration.json#PrimaryDeviceConfiguration" },
    {
      "properties": {
        "roleName": { "type": "string" }
      }
    }
  ]
```

When I reference a definition within the same schema, the error is also gone.

```
  "allOf": [
    { "$ref": "PrimaryDeviceConfiguration.json#PrimaryDeviceConfiguration" },
    { "$ref": "#/$defs/Test" }
  ]
  ...
  "$defs": {
    "Test": {
      "$anchor": "Test",
      "properties": {
        "roleName": { "type": "string" }
      }
    }
  }
```

When I add this property directly to the second schema (`PrimaryDeviceConfiguration.json#PrimaryDeviceConfiguration`), the error is also gone.

```
    "PrimaryDeviceConfiguration": {
      "$anchor": "PrimaryDeviceConfiguration",
      "allOf": [ { "$ref": "DeviceConfiguration.json#DeviceConfiguration" } ],
      "properties": {
        "isPrimaryDevice": { "const": true },
        "roleName": { "type": "string" }
      },
      "required": [ "isPrimaryDevice" ]
    }
  }
```

But, seemingly the property (`roleName`) being defined in the lowest schema (`DeviceConfiguration.json#DeviceConfiguration`) doesn't get picked up.

I would expect the JSON schema specification to dictate these also need to be included, but seemingly this library only recurses one level deep.

justin-tay:
Closes #1123

This fixes the incorrect logic for determining if there is a change in the instance location.

### DIFF
--- a/src/main/java/com/networknt/schema/BaseJsonValidator.java
+++ b/src/main/java/com/networknt/schema/BaseJsonValidator.java
@@ -392,29 +392,18 @@ public abstract class BaseJsonValidator extends ValidationMessageHandler impleme
      */
     protected boolean hasAdjacentKeywordInEvaluationPath(String keyword) {
         JsonSchema schema = getEvaluationParentSchema();
-        boolean checkInstance = false;
-        boolean anchor = false;
-        boolean stop = false;
         while (schema != null) {
             for (JsonValidator validator : schema.getValidators()) {
                 if (keyword.equals(validator.getKeyword())) {
                     return true;
                 }
-                if (checkInstance) {
-                    if ("properties".equals(validator.getKeyword()) || "items".equals(validator.getKeyword())) {
-                        stop = true;
-                    } else if ("$dynamicAnchor".equals(validator.getKeyword()) || "$recursiveAnchor".equals(validator.getKeyword())) {
-                        anchor = true;
-                    }
-                }
             }
-            if (stop && !anchor) {
+            Object element = schema.getEvaluationPath().getElement(-1);
+            if ("properties".equals(element) || "items".equals(element)) {
                 // If there is a change in instance location then return false
                 return false;
             }
             schema = schema.getEvaluationParentSchema();
-            checkInstance = true;
-            anchor = false;
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


stevehu (approved):

Whathecode:
That was quick! Great work, @justin-tay !

PR #1130, opened 2024-11-16
Clean up code

jkosternl:
* Simplify code, add final keyword, use diamond operator, correct English, use lambdas, remove .toString, use .isEmpty(), use .addAll, and more
* Thanks to IntelliJ code analyses feature


stevehu (approved):
It looks like 'sb.append(s, pos, close);' is a syntax for Java 9. We are using Java 11 to build but the output is still Java 8 binary now. Let's see if it works.

stevehu:
Java 8 can compile the code without any issues. Thanks.

PR #1152, opened 2025-02-07
Fix NPE when walking a missing node that will have missing properties

### LINKED ISSUE #1151: NPE during JSON schema walk with validation
We ran into this while migrating to 1.5.5. A JSON schema walk returned an NPE rather than returning validation messages while validating a YAML document that did not have the corresponding properties. Below is the smallest recreate I've found that still fails with an NPE. I've replaced the YAML nodes with an instance of the MissingNode below to simulate the missing object properties.

```java
import com.fasterxml.jackson.databind.node.MissingNode;
import com.networknt.schema.JsonSchema;
import com.networknt.schema.JsonSchemaFactory;
import com.networknt.schema.SpecVersion;

public class SchemaTest {

    public static void main(String[] args) {
        String schemaContents = """
            {
                "type": "object",
                "properties": {
                    "field": {
                    "anyOf": [
                        {
                        "type": "string"
                        }
                    ]
                    }
                }
            }
            """;

        JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
        JsonSchema schema = factory.getSchema(schemaContents);
        schema.walk(MissingNode.getInstance(), true).getValidationMessages();
    }
}
```

Stack trace:
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "com.fasterxml.jackson.databind.JsonNode.getNodeType()" because "node" is null
        at com.networknt.schema.TypeFactory.getValueNodeType(TypeFactory.java:78)
        at com.networknt.schema.utils.JsonNodeUtil.equalsToSchemaType(JsonNodeUtil.java:70)
        at com.networknt.schema.TypeValidator.equalsToSchemaType(TypeValidator.java:50)
        at com.networknt.schema.AnyOfValidator.validate(AnyOfValidator.java:84)
        at com.networknt.schema.AnyOfValidator.walk(AnyOfValidator.java:166)
        at com.networknt.schema.JsonSchema.walk(JsonSchema.java:1278)
        at com.networknt.schema.PropertiesValidator.walkSchema(PropertiesValidator.java:197)
        at com.networknt.schema.PropertiesValidator.validate(PropertiesValidator.java:109)
        at com.networknt.schema.PropertiesValidator.walk(PropertiesValidator.java:133)
        at com.networknt.schema.JsonSchema.walk(JsonSchema.java:1278)
        at com.networknt.schema.JsonSchema.walkAtNodeInternal(JsonSchema.java:1247)
        at com.networknt.schema.JsonSchema.walk(JsonSchema.java:1082)
        at com.networknt.schema.JsonSchema.walk(JsonSchema.java:1162)
       ...
```

justin-tay:
Closes #1151 

Fixes the NPE by having `TypeFactory.getValueNodeType((JsonNode node, SchemaValidatorsConfig config)` handle the case where a `null` was passed in as the node.

I had considered if it should be ensured that the node inputs should always be non null and always replace `null` by `MissingNode` instead. For instance having `PropertiesValidator` set the input as `MissingNode` instead of `null`. But since `JsonNode.get(String fieldName)` returns `null` for missing nodes I wasn't sure it would be appropriate.

### DIFF
--- a/src/main/java/com/networknt/schema/TypeFactory.java
+++ b/src/main/java/com/networknt/schema/TypeFactory.java
@@ -75,6 +75,11 @@ public class TypeFactory {
      * @return the json type
      */
     public static JsonType getValueNodeType(JsonNode node, SchemaValidatorsConfig config) {
+        if (node == null) {
+            // This returns JsonType.UNKNOWN to be consistent with the behavior when
+            // JsonNodeType.MISSING
+            return JsonType.UNKNOWN;
+        }
         JsonNodeType type = node.getNodeType();
         switch (type) {
         case OBJECT:
--- a/src/test/java/com/networknt/schema/JsonWalkTest.java
+++ b/src/test/java/com/networknt/schema/JsonWalkTest.java
@@ -2,6 +2,7 @@ package com.networknt.schema;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.MissingNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.networknt.schema.walk.JsonSchemaWalkListener;
 import com.networknt.schema.walk.WalkEvent;
@@ -108,6 +109,26 @@ class JsonWalkTest {
                 + "          }"
                 + "     }"
                 + "}")));
+    }
+
+    @Test
+    void testWalkMissingNodeWithPropertiesSchema() {
+        String schemaContents = "{\n"
+                + "                \"type\": \"object\",\n"
+                + "                \"properties\": {\n"
+                + "                    \"field\": {\n"
+                + "                    \"anyOf\": [\n"
+                + "                        {\n"
+                + "                        \"type\": \"string\"\n"
+                + "                        }\n"
+                + "                    ]\n"
+                + "                    }\n"
+                + "                }\n"
+                + "            }";
+
+        JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
+        JsonSchema schema = factory.getSchema(schemaContents);
+        schema.walk(MissingNode.getInstance(), true).getValidationMessages();
     }
 
     private InputStream getSchema() {
--- a/src/test/java/com/networknt/schema/JsonWalkTest.java
+++ b/src/test/java/com/networknt/schema/JsonWalkTest.java
@@ -17,6 +17,7 @@ import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.TreeSet;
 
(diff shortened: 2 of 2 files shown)


src/test/java/com/networknt/schema/JsonWalkTest.java
```diff
@@ -110,6 +111,26 @@ void testWalkWithDifferentListeners() throws IOException {
...
+                + "                        \"type\": \"string\"\n"
+                + "                        }\n"
+                + "                    ]\n"
+                + "                    }\n"
+                + "                }\n"
+                + "            }";
+
+        JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
+        JsonSchema schema = factory.getSchema(schemaContents);
+        schema.walk(MissingNode.getInstance(), true).getValidationMessages();
```
jkosternl, line 131: Perhaps assert any valid response from the walk method, or otherwise `assertNonNull`?

src/main/java/com/networknt/schema/TypeFactory.java
```diff
@@ -75,6 +75,11 @@ public static JsonType getSchemaNodeType(JsonNode node) {
      * @return the json type
      */
     public static JsonType getValueNodeType(JsonNode node, SchemaValidatorsConfig config) {
+        if (node == null) {
+            // This returns JsonType.UNKNOWN to be consistent with the behavior when
```
jkosternl, line 79:
```suggestion
            // JsonType.UNKNOWN is consistent with JsonNodeType.MISSING
```
Trying to make it a bit concise?

jkosternl (commented):
Just 2 minor remarks, to take into consideration. Good job!

stevehu (approved):

src/main/java/com/networknt/schema/TypeFactory.java
jkosternl: Resolved without any update, somehow?

justin-tay: Thanks for the suggestion. However your suggestion doesn't convey the intent of the comment that the return result should match the result in the later switch statement for missing node.

jkosternl: Ah I see, thanks for your response 👍🏻 .

stevehu:
@justin-tay The latest commit is not included in the previous PR which is merged so quickly. Could you please submit another PR to include the test case? Thanks.

justin-tay:
@stevehu It looks like the latest commit to add the `assertDoesNotThrow` was included in the merged commit https://github.com/networknt/json-schema-validator/commit/da3865cb4b4155e4321c48998018634c32ce859d so there's nothing more. Thanks!

Issue #1153, opened 2025-02-12
Ethlo and graalvm are not optional when running on the module path.


lscoughlin:
the ethlo.itu library becomes required instead of optional when running on the module path.
They can't seem to get their module-info story together so the project also becomes unusable when running on the library path.

The module dependency should be changed from `require` to `require static` to make the library actually optional for folks running on the module path.

Graal is also marked as required when running on the module path - these dependencies should also be marked as optional by using `required static`.

This patch addresses both.

ethlo:
The module info should be ok with the release of 1.13.0. Thanks for your patience

lscoughlin:
This is a blocker for me for uptake of 1.5.5 ... I don't the ethlo.itu library now that's gotten it's module story better, but I can't have a hard dependency on graalvm and i can't get off the module path.

stevehu:
I am a little bit confused why we need graalvm in the dependency? Is this introduced by one of the dependencies? Thanks.

justin-tay:
`graalvm` is one of the 2 [optional regular expression implementations](https://github.com/networknt/json-schema-validator?tab=readme-ov-file#optional-dependencies) that can be used, the other being `joni`.

I'm not sure why it was explicitly singled out through. There's also the [excludable dependencies](https://github.com/networknt/json-schema-validator?tab=readme-ov-file#excludable-dependencies) `jackson-dataformat-yaml` and `itu`.

I can see that the `module-info` being generated by the moditech plugin looks like the following

```java
module com.networknt.schema {
  requires com.ethlo.time;
  requires com.fasterxml.jackson.annotation;
  requires com.fasterxml.jackson.dataformat.yaml;
  requires org.jruby.jcodings;
  requires org.jruby.joni;
  requires com.fasterxml.jackson.core;
  requires com.fasterxml.jackson.databind;
  requires org.graalvm.sdk;
  requires org.slf4j;
  requires java.base;
  
  exports com.networknt.schema;
  exports com.networknt.schema.annotation;
  exports com.networknt.schema.format;
  exports com.networknt.schema.i18n;
  exports com.networknt.schema.oas;
  exports com.networknt.schema.output;
  exports com.networknt.schema.regex;
  exports com.networknt.schema.resource;
  exports com.networknt.schema.result;
  exports com.networknt.schema.serialization;
  exports com.networknt.schema.serialization.node;
  exports com.networknt.schema.utils;
  exports com.networknt.schema.walk;
}
```

[...]

PR #1154, opened 2025-02-12
makes the ethlo.itu libary optional when running on the module path

lscoughlin:
Address #1153 
makes the ethlo.itu libary optional when running on the module path
- does this by adding 'static' to the requries section of the moditech plugin.
- added optional to the ethlo.itu dependency


stevehu:
I have released 1.5.6 with all the PRs. Please let us know if it works. Thanks.

PR #1157, opened 2025-03-14
Fix for ignoring the result of scheme mapping

TotenTraum:
When loading a schema that uses the "$schema" property set to "http://json-schema.org/schema#" that references the latest version of the specification, the following exception is thrown

``java.io.FileNotFoundException: classpath:schema``

Afterwards I decided to add mapping for such a case, but it still throws the same exception
``` java
final VersionFlag versionFlag = VersionFlag.V7;
final String schemaIri = JsonSchemaFactory.checkVersion(versionFlag).getInstance().getIri();

final JsonSchemaFactory jsonSchemaFactory = JsonSchemaFactory.getInstance(versionFlag, builder ->
     builder.schemaMappers(schemaMappers -> schemaMappers.mappings(Map.of("http://json-schema.org/schema", schemaIri)))
);
```

It turned out that the code of the DefaultSchemaLoader class ignores the mapping result
``` java
...
AbsoluteIri mappedResult = absoluteIri; <-----
for (SchemaMapper mapper : schemaMappers) {
    AbsoluteIri mapped = mapper.map(mappedResult);
    if (mapped != null) {
         mappedResult = mapped;
       }
    }
    AbsoluteIri mapped = META_SCHEMA_MAPPER.map(absoluteIri); <----
    if (mapped != null) {
        mappedResult = mapped;
    }
...
```

### DIFF
--- a/src/main/java/com/networknt/schema/resource/DefaultSchemaLoader.java
+++ b/src/main/java/com/networknt/schema/resource/DefaultSchemaLoader.java
@@ -51,7 +51,7 @@ public class DefaultSchemaLoader implements SchemaLoader {
                 mappedResult = mapped;
             }
         }
-        AbsoluteIri mapped = META_SCHEMA_MAPPER.map(absoluteIri);
+        AbsoluteIri mapped = META_SCHEMA_MAPPER.map(mappedResult);
         if (mapped != null) {
             mappedResult = mapped;
         }


justin-tay:
Just a note that the use of http://json-schema.org/schema# to refer to the latest version is deprecated and should not be used. See https://json-schema.org/understanding-json-schema/reference/schema. Dialects are not backward compatible so you may get unexpected results.

That said you need to add a unit test for the pull request for this to be reviewed if this is still something you want.

PR #1160, opened 2025-03-27
Refactored Code to Improve Maintainability and Modularity 

keshvi00:
This pull request introduces six refactoring improvements to address implementation and design smells identified in the codebase. The goal is to enhance readability, modularity, and maintainability without altering core functionality. Below is a summary of the changes:

Set 1: Implementation Smell Fixes

1. Decomposed Conditional (EmailValidator - isValidDomain)
Smell: Complex conditional logic made it difficult to read and maintain.
Solution: Extracted conditions into separate methods, making the logic clearer and improving readability.

2. Extracted Method (BaseJsonValidator - registerAndMergeDiscriminator)
Smell: Large method (registerAndMergeDiscriminator) contained multiple responsibilities.
Solution: Extracted helper methods to improve readability and single responsibility.

3. Renamed Long Identifier (AbsoluteIris - unicodeToASCII)
Smell: Variable IDN_TOASCII_PRESERVES_TRAILING_DOTS was too long.
Solution: Shortened it to IDN_KEEPS_TRAILING_DOTS while preserving its context.

Set 2: Design Smell Fixes

1. Moved Method (BaseJsonValidator - checkDiscriminatorMatch, registerAndMergeDiscriminator)
Smell: Core validation logic was mixed with discriminator handling, making BaseJsonValidator too large (400 lines).
Solution: Moved discriminator-specific methods to DiscriminatorContext, reducing complexity.

2. Replaced Conditional with Polymorphism (BaseJsonValidator - atRoot)
Smell: The atRoot method used an if-else chain based on PathType.
Solution: Introduced a PathStrategy interface with multiple implementations (LegacyPathStrategy, JsonPointerPathStrategy, JsonPathStrategy, DefaultPathStrategy).

3. Pushed Down Method (ContainsValidator - validate)
Smell: ContainsValidator mixed validation logic with annotation collection.
Solution: Moved annotation-specific logic to a new subclass, AnnotatedContainsValidator.

### DIFF
--- a/src/main/java/com/networknt/org/apache/commons/validator/routines/EmailValidator.java
+++ b/src/main/java/com/networknt/org/apache/commons/validator/routines/EmailValidator.java
@@ -206,7 +206,15 @@ public class EmailValidator implements Serializable {
         }
         // Domain is symbolic name
         if (allowTld) {
-            return domainValidator.isValid(domain) || !domain.startsWith(".") && domainValidator.isValidTld(domain);
+            if (domainValidator.isValid(domain)) {
+                return true;
+            }
+
+            if (!domain.startsWith(".") && domainValidator.isValidTld(domain)) {
+                return true;
+            }
+
+            return false;
         }
         return domainValidator.isValid(domain);
     }
--- a/src/main/java/com/networknt/schema/BaseJsonValidator.java
+++ b/src/main/java/com/networknt/schema/BaseJsonValidator.java
@@ -182,6 +182,24 @@ public abstract class BaseJsonValidator extends ValidationMessageHandler impleme
         }
     }
 
+    private static void updateMappingWithFieldsCheck(Iterator<Map.Entry<String, JsonNode>> fieldsToAdd,
+                                                    ObjectNode mappingOnContextDiscriminator, JsonNodePath instanceLocation) {
+        while (fieldsToAdd.hasNext()) {
+            final Map.Entry<String, JsonNode> fieldToAdd = fieldsToAdd.next();
+            final String mappingKeyToAdd = fieldToAdd.getKey();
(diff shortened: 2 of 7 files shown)


justin-tay:
Not keen on this, plus you removed some test cases, and looks like you are refactoring for the sake of refactoring. If you want something to work on an automated benchmark that runs on each pull request would be nice.

src/main/java/com/networknt/org/apache/commons/validator/routines/EmailValidator.java
```diff
@@ -206,7 +206,15 @@ protected boolean isValidDomain(final String domain) {
...
-            return domainValidator.isValid(domain) || !domain.startsWith(".") && domainValidator.isValidTld(domain);
+            if (domainValidator.isValid(domain)) {
+                return true;
+            }
+
+            if (!domain.startsWith(".") && domainValidator.isValidTld(domain)) {
+                return true;
+            }
+
+            return false;
```
jkosternl, line 217: These changes are not needed, because the previous impl was easy to read already. Complexity was not high enough to say: yes, this is much better.

justin-tay: In any case these sets of code are supposed to be copied wholesale from upstream https://github.com/apache/commons-validator/blob/master/src/main/java/org/apache/commons/validator/routines/EmailValidator.java with just a change in package name so as not to pull the whole dependency in.

jkosternl: Ah yes, of course, you're right. I didn't even notice that the package is from the external repo of `commons-validator`. Keeping in sync with the library is a good idea, to not get deviations and invent the wheel (and possibly bugs) twice.

Issue #1170, opened 2025-06-03
Missing RFC valid TLDs for email format


egervais7:
The [GENERIC_TLDS](https://github.com/networknt/json-schema-validator/blob/ffe4ca0e5b6c8cedf78b418106e900998dfdbbd3/src/main/java/com/networknt/org/apache/commons/validator/routines/DomainValidator.java#L181-L1448) is missing domains listed as valid for testing [here](https://www.rfc-editor.org/rfc/rfc2606.html?utm_source=chatgpt.com#section-2)

These include:

```
.test
.example
.invalid
.localhost
```

justin-tay:
Not sure what your use case for these changes are. These changes will affect those using this for production use cases as now obviously invalid testing emails will now be accepted.

Also these codes are copied from https://commons.apache.org/proper/commons-validator/ so with no unit tests these changes will likely be gone the next time these codes get updated from upstream.

It might make more sense to get this merged upstream first instead of here.

PR #1171, opened 2025-06-03
add TLDs for testing

egervais7:
To resolve [issue#1170](https://github.com/networknt/json-schema-validator/issues/1170)

Adding TLDs for testing per RFC documentation seen [here](https://www.rfc-editor.org/rfc/rfc2606.html?utm_source=chatgpt.com#section-2)

Which notes:

```
To safely satisfy these needs, four domain names are reserved as
   listed and described below.

                   .test
                .example
                .invalid
              .localhost

      ".test" is recommended for use in testing of current or new DNS
      related code.

      ".example" is recommended for use in documentation or as examples.

      ".invalid" is intended for use in online construction of domain
      names that are sure to be invalid and which it is obvious at a
      glance are invalid.

      The ".localhost" TLD has traditionally been statically defined in
      host DNS implementations as having an A record pointing to the
      loop back IP address and is reserved for such use.  Any other use
      would conflict with widely deployed code which assumes this use.
```

### DIFF
--- a/src/main/java/com/networknt/org/apache/commons/validator/routines/DomainValidator.java
+++ b/src/main/java/com/networknt/org/apache/commons/validator/routines/DomainValidator.java
@@ -812,7 +812,7 @@ public class DomainValidator implements Serializable {
         "llp", // llp Dot Registry LLC
         "loan", // loan dot Loan Limited
         "loans", // loans June Woods, LLC
-        "localhost", // RFC reserved for DNC with A record pointing to loop back IP address
+        "localhost", // RFC reserved for DNS with A record pointing to loop back IP address
         "locker", // locker Dish DBS Corporation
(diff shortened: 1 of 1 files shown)


justin-tay:
Not sure what your use case for these changes are. These changes will affect those using this for production use cases as now obviously invalid testing emails will now be accepted.

Also these codes are copied from https://commons.apache.org/proper/commons-validator/ so with no unit tests these changes will likely be gone the next time these codes get updated from upstream.

It might make more sense to get this merged upstream first instead of here.

justin-tay:
Does the address actually resolve? Still not really clear why you would want this to pass in a production setting.

One of the issues in any case is that the apache commons email validator implementation isn't particularly spec compliant to start with.

PR #1220, opened 2025-12-17
fixes #1206 move the SpecificationVersionDetector to main from test

### LINKED ISSUE #1206: Why was SpecificationVersionDetector moved to test in 2.0?
Why was `SpecificationVersionDetector` moved to the test module in 2.0?
Previously, this class was part of the main source set and could be used by other components at runtime. Moving it to the test scope may indicate it’s now considered internal or only used for validation purposes. It would be helpful to clarify whether this change affects any public API or integration points.

stevehu:


justin-tay (commented):
I actually don't think this is a good idea particularly as some of what it is doing like determining the version based on the file path seems rather odd to be exposing to users. 

I'm not sure if the users really know what they are doing as the dialect is already automatically detected from `$schema`. A lot of people use their own `ObjectMapper` to parse it to `JsonNode`, and this is rather unnecessary, particularly if you want to have line numbers as the library needs to use it's own to do so.

I don't mind if something like `public static Optional<SpecificationVersion> fromSchemaNode(JsonNode)` was added to `SpecificationVersion` but would like to know exactly what their use case is and if there is a better way to solve their use case.

justin-tay:
While the approach sounds reasonable, the upgrade to 2.x was already a rather large breaking change. Without resolving the actual underlying issue, this will just come up again for 3.x.

PR #1221, opened 2025-12-18
Add method to SpecificationVersion to determine from schema node

justin-tay:
Closes #1220


sschuberth:
> Closes #1220

Actually, I believe this relates more the underlying issue #1206 than to PR #1220.

sschuberth:
And is there a chance to get a 2.x patch release with this change still? (FYI @nscuro.)

PR #1232, opened 2026-03-08
oneOf and discriminator value mismatch doesnot fail

stevehu:


stevehu:
Based on the comment from @justin-tay, we should drop this PR as the current behaviour is correct.

PR #1247, opened 2026-05-26
fixes #1246 Coercion bug

### LINKED ISSUE #1246: Coercion bug
After update from Jackson 2 to Jackson 3, and respectively updating from `com.networknt:json-schema-validator:2.0.1` to `3.0.2`, the following issue occurred and one unit test started failing.

Exception:
```java
tools.jackson.databind.exc.JsonNodeException: 'ObjectNode' method `asString()` cannot coerce value {...(1 properties}] to `java.lang.String`: value type not coercible
```

Full stack trace: [stack_trace.txt](https://github.com/user-attachments/files/28271935/stack_trace.txt)
Repository with example test to reproduce the issue: [repo](https://github.com/Ivan-Shaml/reproduce-coercion-bug)

stevehu:


src/main/java/com/networknt/schema/dialect/Dialect.java
```diff
@@ -414,10 +414,10 @@ public String readDynamicAnchor(JsonNode schemaNode) {
         return null;
     }
 
-    private static String readText(JsonNode node, String field) {
-        JsonNode fieldNode = node.get(field);
-        return fieldNode == null ? null : fieldNode.asString();
-    }
+    private static String readText(JsonNode node, String field) {
+        JsonNode fieldNode = node.get(field);
+        return fieldNode != null && fieldNode.isString() ? fieldNode.asString() : null;
```
justin-tay: Since the issue is that it shouldn't be coercing then these changes are confusing as `asString()` does coercing but `isString()` ensures it's a `StringNode` so the intent isn't clear.

More appropriate to use the following instead
```java
    private static String readText(JsonNode node, String field) {
        JsonNode fieldNode = node.get(field);
        return fieldNode == null ? null : fieldNode.stringValue(null);
    }
```

justin-tay (changes requested):

src/main/java/com/networknt/schema/dialect/Dialect.java
stevehu: Makes sense. I have made the code change based on your recommendation. Thanks a lot.

PR #1250, opened 2026-06-10
fixes #1174 TextNodes as schema seem to validate any value

### LINKED ISSUE #1174: TextNodes as schema seem to validate any value
Hi, first of, thanks for all your work you put into this library.

**What is the issue**

Given this test, it seems that if I (e.g. accidentally) put a TextNode into a schema, this schema will validate anything (I tested integers and strings).

```java
@Test
void testSchemaStringValidation() {
    JsonNode schemaNode = new TextNode("false");
    JsonSchema schema = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4).getSchema(schemaNode);
    Set<ValidationMessage> errors = schema.validate(new IntNode(42));
    assertTrue(errors.isEmpty(), "Schema should validate any string, but got errors: " + errors);
}
```

**Why is this an issue?**

I was accidentally not parsing a JSON string that I got from an API. I read the response JSON, but forgot to parse the JSON string given in one of its values. The result was that I spent hours diving into what is wrong with this schema (but not with others I got from another API), checking versions, wondering if $refs are the issue... until I saw in the debugger that my JsonNode was a TextNode instance and tried out if this had an effect. I think there is no clue on this in the API docs or so, but to be honest, while debugging this was not the kind of detail I was actively looking for. So, maybe I just did not see it.

**What would I expect?**

I would find it helpful if the library would only accept ObjectNode and maybe a variant with boolean nodes or so. Or it could throw a specialized error on validation, or an exception on schema creation. I think the minimum would be a note in the Javadoc, or make the note more present if it is already somewhere.

stevehu:


stevehu:
Viewed spec.yaml:39-84
Ran command: `curl -sL https://patch-diff.githubusercontent.com/raw/networknt/json-schema-validator/pull/1250.diff`
Ran command: `curl -sL https://patch-diff.githubusercontent.com/raw/networknt/json-schema-validator/pull/1250.diff | head -n 100`

This is an excellent pull request for specification compliance and strictness. Here is my review of the changes in PR #1250:

### 1. Strict Schema Node Type Validation
The most significant change is the introduction of strict type validation for schema nodes (`validateSchemaNodeType`). Previously, the validator might have silently ignored or swallowed invalid JSON nodes that were mistakenly passed as schemas or sub-schemas (such as strings or arrays). Now, it strictly enforces that **all** schemas must be either a JSON `object` or a `boolean`. 

This strictness is so effective that it actually caught several malformed schemas in your own test suite! For example:
- In `issue656.json`, `"frequency": "integer"` was corrected to `"frequency": {"type": "integer"}`.
- In `UnknownMetaSchemaTest.java`, the `"required": ["data"]` array was mistakenly nested *inside* the `properties` object. Because the new logic validates sub-schemas, it correctly realized that an array is not a valid JSON schema object and forced the test to be corrected.

### 2. Specification-Aware Boolean Schemas
The PR correctly implements boolean schema support based on the JSON Schema draft version. By introducing `supportsBooleanSchema`, it ensures that boolean schemas (`true` / `false`) are accepted in Draft 6 and newer, but correctly rejected in Draft 4 (which strictly required schemas to be objects).

### 3. Graceful Fragment Resolution
The overloaded `loadSchema(schemaUri, validateLoadedSchema)` method is a smart addition. It allows the registry to download a raw JSON document and resolve a specific fragment (e.g., `url#/my-schema`) without prematurely validating the root document as a schema. This correctly handles edge cases where a schema is embedded inside a larger, non-schema JSON payload.

### Conclusion
This PR massively improves the robustness of the validator. It prevents users from writing silent bugs in their JSON schemas (like typing `"id": "string"` instead of `{"type": "string"}`) by failing fast and throwing a clear `SchemaException`. The code is clean, the fallback logic for dialect versions is correct, and the accompanying test fixes prove that the new validation is working exactly as intended. 

Everything looks great and ready to merge!

PR #1253, opened 2026-06-20
fixes #1252 problem with additionalProperties schema in jsconfig schema

### LINKED ISSUE #1252: problem with additionalProperties schema in jsconfig schema
Given the following schema https://github.com/katydid/validator-jsonschema-benchmarks/blob/main/schemas/jsconfig-rmUniqueItems/schema.json

Where we see this subschema
```json
"paths": {
              "description": "Specify a set of entries that re-map imports to additional lookup locations.",
              "type": "object",
              "additionalProperties": {
                "type": "array",
                "items": {
                  "type": "string",
                  "description": "Path mapping to be computed relative to baseUrl option."
                }
              },

            },
```

It wants the fields to all have arrays of strings

example valid input
```json
{"compilerOptions": {"paths": {"#utils/*": ["./src/utils/*"], "#controllers/*": ["./src/controllers/*"], "#services/*": ["./src/services/*"], "#routes/*": ["./src/routes/*"], "#models/*": ["./src/models/*"], "#validations/*": ["./src/validations/*"], "#middlewares/*": ["./src/middlewares/*"]}, "target": "ESNext"}, "exclude": ["node_modules"]}
```

I have mutated into for a mutation test
```json
{"compilerOptions":{"paths":{"#controllers/*":["./src/controllers/*"],"#middlewares/*":["./src/middlewares/*"],"#models/*":["./src/models/*"],"#routes/*":[{"3k2aNc8lWP4Zda3sGft31yi898z4N4x63p4GtLAGDe1nn9r7ne3Uq7d68":null,"8E9D276L01a46bKiY03":false,"jLS13ILLJ0a9v71C80wMsW6Q0hNAqtViS61uKN4iWRC3B3tT7K":false}],"#services/*":["./src/services/*"],"#utils/*":["./src/utils/*"],"#validations/*":["./src/validations/*"]},"target":"ESNext"},"exclude":["node_modules"]}
```

We replaced 
```
"#routes/*": ["..."]
```
With
```
"#routes/*":[{...}]
```

Now the field is an array of object and not array of string, which is invalid

But this library is returning that it is valid
version: 3.0.3

stevehu:


justin-tay (approved):
Think the lines for skipping the context items were there since the initial commit and was never really sure what use case they were for. Don't really see an issue with removing it.

PR #1254, opened 2026-06-21
Backport master fixes

stevehu:


stevehu:
Findings:

  - Low: json-schema-validator/src/main/java/com/networknt/schema/SchemaRegistry.java:290 has one real trailing-whitespace issue. git -c core.whitespace=cr-at-eol diff --check
    origin/2.x...HEAD reports only this line after accounting for the repo’s CRLF-heavy Java files.

  No blocking correctness findings. The backported schema validation, $ref fragment handling, additionalProperties # key fix, and regex context propagation look consistent with the related
  master fixes.

PR #1258, opened 2026-06-25
test: re-enable and strengthen recursive $ref SelfRefTest

koteshyelamati:
SelfRefTest has been @Disabled since 2016 with the note that a recursive $ref caused a StackOverflowError. That has long since been fixed, but the test stayed disabled and only printed to stdout, so the recursive-$ref path had no coverage.

This re-enables it and replaces the println with real assertions: the self-referential "tree" schema loads without StackOverflow, a valid nested tree validates with no errors, and a nested branch missing the required "value" is reported (proving validation follows the recursive $ref). Verified locally with mvn -Dtest=SelfRefTest test.

### DIFF
--- a/src/test/java/com/networknt/schema/SelfRefTest.java
+++ b/src/test/java/com/networknt/schema/SelfRefTest.java
@@ -16,17 +16,49 @@
 
 package com.networknt.schema;
 
-import org.junit.jupiter.api.Disabled;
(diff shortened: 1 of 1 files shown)


stevehu (approved):

stevehu:
@koteshyelamati Thanks for your help.

Issue #1260, opened 2026-06-29
Problem with allOf in v3.0.5


gjvoosten:
[EDIT: replaced #1252 with #1250 as that PR seems to have introduced the issue.]

The fix in #1250 appears to introduce a new problem, as I now get an error such as:

> Caused by: com.networknt.schema.SchemaException: Schema at […]/allOf must be object or boolean but was ARRAY
> at com.networknt.schema.SchemaRegistry.validateSchemaNodeType(SchemaRegistry.java:542)
> […]

but [the spec](https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.1.1) states:

_"This keyword's value MUST be a non-empty array. Each item of the array MUST be a valid JSON Schema."_

With 3.0.4 this error did not happen.

I've created #1261 that tests a reduced schema against some valid and invalid files. When run against v3.0.4 (commit https://github.com/networknt/json-schema-validator/commit/2276a17e117c8977eacac2dcef04908939463bed ), these tests all pass (as expected), but against v3.0.4 plus PR #1250 (commit https://github.com/networknt/json-schema-validator/commit/e5f6dc1c35ebadc63808136d19be002ca281e2ab ), as well as against current master, the test's `setup()` fails with:

> [ERROR] Errors: 
> [ERROR]   Issue1260Test.setup:19->BaseJsonSchemaValidatorTest.getJsonSchemaFromClasspath:56->BaseJsonSchemaValidatorTest.getJsonSchemaFromClasspath:63 » Schema Schema at #/$defs/assessmentDef/dependentSchemas/recurrence/oneOf/2/properties/questions/properties/allOf must be object or boolean but was ARRAY

gjvoosten:
Additional tests were added through #1261.

stevehu:
@gjvoosten Thanks for your help. These test cases clarifies a lot of confusions.

PR #1261, opened 2026-06-29
Add schema tests (#1260)

gjvoosten:
These tests pass against 3.0.4 as well as 3.0.5.

### DIFF
--- a/src/test/java/com/networknt/schema/Issue1260Test.java
+++ b/src/test/java/com/networknt/schema/Issue1260Test.java
@@ -1,0 +1,41 @@
+package com.networknt.schema;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class Issue1260Test extends BaseJsonSchemaValidatorTest {
+
+    private static final String RESOURCE_PREFIX = "issues/1260/";
+    private static Schema schema;
+
+    @BeforeAll
+    static void setup() {
+        schema = getJsonSchemaFromClasspath(resource("schema.json"), SpecificationVersion.DRAFT_2020_12);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"valid1.json", "valid2.json", "valid3.json"})
+    void testNoErrorsForValidInput(String validJson) throws IOException {
+        final var node = getJsonNodeFromClasspath(resource(validJson));
+        final var errors = schema.validate(node);
+        assertTrue(errors.isEmpty(), "No validation errors for valid input");
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"invalid1.json", "invalid2.json", "invalid3.json"})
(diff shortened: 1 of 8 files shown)


stevehu:
Yes, the root cause of the issue described in GitHub Issue #1260 https://github.com/networknt/json-schema-validator/issues/1260 is a set of syntax errors in the
  user's JSON Schema that were previously ignored silently in  v3.0.4  but are correctly rejected in  v3.0.5 .
  ──────
  ### Root Cause Analysis

  #### 1. Misplacement of  allOf  inside  properties 

  In the user's schema:

    "questions": {
      "properties": {
        "allOf": [
          { "$ref": "#/$defs/questions" },
          { "assessmentDate": "#/$defs/question" },
          { "expirationDate": "#/$defs/question" }
        ]
      }
    }
    
  Because  "allOf"  is defined inside the  "properties"  block of  "questions" , the validator treats  "allOf"  as a property name rather than the JSON Schema  allOf 
  keyword.

  According to the JSON Schema specification, the schema for any property under  "properties"  must be a valid JSON Schema (which can only be a JSON object or a
  boolean). Because the value of the  "allOf"  property here is an array ( [...] ), the validator throws:

  │  SchemaException: Schema at […]/allOf must be object or boolean but was ARRAY 

  #### 2. Invalid Subschema syntax

  Furthermore, inside the array, they wrote:

    { "assessmentDate": "#/$defs/question" }
    
  This is also invalid JSON Schema syntax (it is an object using a custom property key  assessmentDate  with a string value, rather than a proper  {"$ref": ...} 
  schema object).
  ──────
  ### Why it worked in  v3.0.4  but fails in  v3.0.5 

  • In  v3.0.4 : The library did not validate loaded schema node types at compilation/load time. Because the input JSON never actually contained a property named 
  "allOf" , the validator never had to run validation on it, and the schema compiled without crashing.
  • In  v3.0.5  (introduced by PR #1250): The library added strict compile-time validation of loaded schemas in  SchemaRegistry.validateSchemaNodeType . This ensures
  that schemas are structurally valid when compiled. Since the value of the property  "allOf"  is an array, the validator correctly rejects it with a  SchemaException
  during initialization.
  ──────
  ### How to Fix the Schema

  The user's schema should be corrected to place  allOf  as a sibling keyword of  properties  rather than nesting it inside, and specify the subschemas correctly:

     "questions": {
    -  "properties": {
    -    "allOf": [
    -      {
    -        "$ref": "#/$defs/questions"
    -      },
    -      {
    -        "assessmentDate": "#/$defs/question"
    -      },
    -      {
    -        "expirationDate": "#/$defs/question"
    -      }
    -    ]
    -  },
    +  "allOf": [
    +    {
    +      "$ref": "#/$defs/questions"
    +    },
    +    {
    +      "properties": {
    +        "assessmentDate": {
    +          "$ref": "#/$defs/question"
    +        }
    +      }
    +    },
    +    {
    +      "properties": {
    +        "expirationDate": {
    +          "$ref": "#/$defs/question"
    +        }
    +      }
    +    }
    +  ],
       "required": [
         "assessmentDate"
       ]
     }

gjvoosten:
@stevehu Thanks for your thorough analysis. I must admit that that specific part of the JSON Schema draft was always a bit foggy. I've updated the schema in the PR, and the tests now pass against 3.0.4 as well as 3.0.5. Might be worth including these to avoid future regressions. Merge, or close, at your discretion.

stevehu (approved):

PR #1265, opened 2026-07-07
Fall back to the library class loader for classpath resources

### LINKED ISSUE #1264: Bundled meta-schemas fail to load under OSGi (TCCL cannot see the library's resources)
Loading the bundled 2020-12 meta-schema fails under OSGi:

    SchemaException: Failed to load json schema from https://json-schema.org/draft/2020-12/schema
    Caused by: FileNotFoundException: classpath:draft/2020-12/schema

`ClasspathResourceLoader` resolves `classpath:` IRIs through the thread context class loader, falling back to the library's own class loader only when the TCCL is null. Under OSGi the TCCL is non-null but cannot see resources inside this jar, so the lookup throws. I hit this while developing an OSGi application; it was also reported independently downstream in modelcontextprotocol/java-sdk#992: the MCP SDK loads the meta-schema in its validator constructor, so every server fails at startup.

Configuring a resource loader doesn't help: `SchemaLoader.getSchemaResource()` consults the static classpath loader before any configured loaders, so `classpath:` IRIs never reach them. The only seam is the classpath loader itself.

This is a regression from 1.x. `ClasspathURLStreamHandler` chained TCCL → own class loader → system class loader; the chain was not carried over in the schema retrieval refactor (#931).

Reproduces without an OSGi runtime:

```java
Thread.currentThread().setContextClassLoader(new ClassLoader(null) {
    @Override public InputStream getResourceAsStream(String name) { return null; }
});
SchemaRegistry.withDefaultDialect(Dialects.getDraft202012())
    .getSchema(SchemaLocation.of("https://json-schema.org/draft/2020-12/schema"));
// FileNotFoundException: classpath:draft/2020-12/schema
```

Affects both 2.x and 3.x — the loader is identical. PRs restoring the 1.x-style fallback to follow, for master and 2.x.

joaodinissf:
Closes #1264

Retries with the library's own class loader after the thread context class loader misses. The TCCL still wins on a hit; a custom class loader source is never overridden. The test covers the fallback, TCCL precedence, and custom-source passthrough with stub class loaders — no new test dependencies. (Also verified in an embedded Felix framework; that harness is not committed.)

### DIFF
--- a/src/main/java/com/networknt/schema/resource/ClasspathResourceLoader.java
+++ b/src/main/java/com/networknt/schema/resource/ClasspathResourceLoader.java
@@ -35,11 +35,14 @@ public class ClasspathResourceLoader implements ResourceLoader {
 
     private final Supplier<ClassLoader> classLoaderSource;
 
+    /** Whether to fall back to this library's own class loader when the default source misses. */
+    private final boolean fallbackToOwnClassLoader;
+
     /**
      * Constructor.
      */
     public ClasspathResourceLoader() {
-        this(ClasspathResourceLoader::getClassLoader);
+        this(ClasspathResourceLoader::getClassLoader, true);
     }
 
     /**
@@ -48,7 +51,12 @@ public class ClasspathResourceLoader implements ResourceLoader {
      * @param classLoaderSource the class loader source
      */
     public ClasspathResourceLoader(Supplier<ClassLoader> classLoaderSource) {
+        this(classLoaderSource, false);
+    }
+
+    private ClasspathResourceLoader(Supplier<ClassLoader> classLoaderSource, boolean fallbackToOwnClassLoader) {
         this.classLoaderSource = classLoaderSource;
+        this.fallbackToOwnClassLoader = fallbackToOwnClassLoader;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


stevehu:
No blocking findings.

  Risk: Medium-low. The code change is tiny, but ClasspathResourceLoader#getResource sits under schema loading and $ref/meta-schema resolution. GitNexus shows direct callers in
  SchemaLoader.getSchemaResource, SchemaLoader.getClasspathResource, and Schema.getInputResource, with the affected flow centered on schema loading.

  Coverage looks good for the intended behavior:

  - fallback when TCCL cannot see bundled resources: json-schema-validator/src/test/java/com/networknt/schema/resource/ClasspathResourceLoaderTcclTest.java:66
  - TCCL still wins when it can resolve the resource: json-schema-validator/src/test/java/com/networknt/schema/resource/ClasspathResourceLoaderTcclTest.java:74
  - explicit custom classloader source is not overridden: json-schema-validator/src/test/java/com/networknt/schema/resource/ClasspathResourceLoaderTcclTest.java:95

  Areas to improve, non-blocking:

  - There is no committed real OSGi/Felix integration test. The stub classloader regression test is reasonable and cheap, but the actual bundle classloader behavior remains manually
    verified.

  - The fallback uses ResourceLoader.class.getClassLoader() at json-schema-validator/src/main/java/com/networknt/schema/resource/ClasspathResourceLoader.java:85. That should be equivalent
    in this package, but ClasspathResourceLoader.class.getClassLoader() would make the intent slightly more local and explicit.

stevehu:
It appears that PR #1265 contains the **exact same code changes** as PR #1266 (both modifying `ClasspathResourceLoader.java` and adding `ClasspathResourceLoaderTcclTest.java` with 123 insertions). They likely represent duplicate PRs opened by the author (perhaps accidentally or against different branches).

Since the codebase changes are identical, **my previous review applies in its entirety to this PR as well.**

To summarize:
* **The fix is correct**: It safely falls back to the library's own class loader when the Thread Context Class Loader (TCCL) fails to find bundled resources (which is the root cause of the OSGi issue).
* **It preserves backward compatibility**: Custom class loader suppliers are respected and do not trigger the fallback unexpectedly.
* **The tests are solid**: They cover both the blind TCCL case and cases where TCCL and the library class loader compete.

**Status: Approved / LGTM (Looks Good To Me)**. 

You can safely merge whichever of the two PRs (#1265 or #1266) targets the branch you intend to update, and close the other as a duplicate!

PR #1267, opened 2026-07-22
Reject non-ASCII whitespace in email format

### LINKED ISSUE #1164: email format allows nbsp character
A project I work on uses
```
"format": "email"
```
in the schema, and we had an email pass validation with a leading nbsp character, i.e. 
```
" name@email.com"
```
Since most other whitespace in the email results in validation errors, I believe this should too. The [regex](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/org/apache/commons/validator/routines/EmailValidator.java#L40) excludes `\s` characters, but nbsp characters are not included in that.

el-psy-kongroo-d:
## Problem

`{ "format": "email" }` accepts values containing a non-breaking space (U+00A0)
or other non-ASCII whitespace. For example, an email with a leading U+00A0 is
treated as valid, even though the same address with a regular leading space is
correctly rejected.

The underlying `EmailValidator` regex only excludes ASCII whitespace, so
non-ASCII whitespace slips through.

## Fix

`EmailFormat` now rejects any value containing a non-ASCII whitespace character
before delegating to the validator. `Character.isWhitespace` deliberately
excludes U+00A0, so `Character.isSpaceChar` is also checked.

The change only affects values that contain non-ASCII whitespace; all-ASCII
emails — including quoted local parts with spaces such as
`"joe bloggs"@example.com` — are unchanged.

Following the existing per-format test convention (`TimeFormatTest`,
`UriFormatTest`, …), this also adds `EmailFormatTest`.

## Note on idn-email

`IdnEmailFormat` (idn-email) shares the same delegation and therefore the same
behavior, but is intentionally left out of scope here since this issue concerns
`email`. Happy to address it in a follow-up if desired.

## Testing

- New `EmailFormatTest` covers a plainly valid email and the reported
  non-breaking space case.
- The full test suite passes locally, including the official JSON Schema Test
  Suite email cases (e.g. the valid quoted local part `"joe bloggs"@example.com`).

Closes #1164

### DIFF
--- a/src/main/java/com/networknt/schema/format/EmailFormat.java
+++ b/src/main/java/com/networknt/schema/format/EmailFormat.java
@@ -35,7 +35,25 @@ public class EmailFormat implements Format {
 
     @Override
     public boolean matches(ExecutionContext executionContext, String value) {
+        if (containsNonAsciiWhitespace(value)) {
+            return false;
+        }
         return this.emailValidator.isValid(value);
+    }
+
+    /**
+     * Whether {@code value} contains a non-ASCII whitespace character such as
+     * {@code U+00A0} NON-BREAKING SPACE. The underlying validator only excludes
+     * ASCII whitespace, so such characters would otherwise be accepted in an
+     * email address. Note that {@link Character#isWhitespace(char)} deliberately
+     * excludes {@code U+00A0}, hence the additional
+     * {@link Character#isSpaceChar(char)} check.
+     */
+    private static boolean containsNonAsciiWhitespace(String value) {
+        if (value == null) {
+            return false;
+        }
+        return value.codePoints().anyMatch(ch -> ch > 0x7F && (Character.isWhitespace(ch) || Character.isSpaceChar(ch)));
     }
     
     @Override
--- a/src/test/java/com/networknt/schema/format/EmailFormatTest.java
+++ b/src/test/java/com/networknt/schema/format/EmailFormatTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.networknt.schema.format;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import com.networknt.schema.Error;
+import com.networknt.schema.InputFormat;
+import com.networknt.schema.Schema;
+import com.networknt.schema.SchemaRegistry;
+import com.networknt.schema.SpecificationVersion;
+
+class EmailFormatTest {
+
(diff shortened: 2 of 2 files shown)


stevehu:
Code Review — PR #1267: Reject non-ASCII whitespace in email format

Overview

Fixes issue #1164: {"format":"email"} accepts addresses containing non-ASCII whitespace (e.g. a leading U+00A0 NON-BREAKING SPACE), while the same address with a regular ASCII space is correctly rejected. The root cause is that the underlying EmailValidator regex only excludes ASCII whitespace. The PR adds a pre-check in EmailFormat.matches() that rejects any value containing a non-ASCII whitespace code point before delegating, and adds a new EmailFormatTest. 85 additions, 2 files, no deletions.

Correctness ✅

- The guard logic is sound:
value.codePoints().anyMatch(ch -> ch > 0x7F && (Character.isWhitespace(ch) || Character.isSpaceChar(ch)))
  - ch > 0x7F scopes the check to non-ASCII, so all-ASCII emails are completely untouched — including the quoted-local-part case "joe bloggs"@example.com, whose space is 0x20 and is skipped. This directly preserves the JSON Schema Test Suite positive cases.
  - The isWhitespace || isSpaceChar combination is the correct idiom: Character.isWhitespace deliberately excludes U+00A0/U+2007/U+202F, so isSpaceChar (Unicode Zs/Zl/Zp) is needed to catch them. The Javadoc explains this subtlety well.
  - codePoints() (int-based) handles the full range correctly; null returns false and preserves the existing isValid(null) behavior. Good.
- I confirmed against the base branch that every API the test uses (Error, InputFormat, Schema, SchemaRegistry.withDefaultDialect, SpecificationVersion.DRAFT_2020_12) exists — the test will compile.

Conventions 👍

- EmailFormatTest mirrors UriFormatTest almost line-for-line (schema string, withDefaultDialect(...).getSchema(...), formatAssertionsEnabled(true)) — exactly the stated convention. Building the NBSP from Character.toChars(0x00A0) rather than pasting an invisible character into source is a nice, reviewable touch.
- Minor nit: the new file's license header reads Copyright (c) 2025 the original author or authors, whereas the existing source files (e.g. EmailFormat.java) use Copyright (c) 2016 Network New Technologies Inc.. Cosmetic, but worth matching the repo's prevailing header for consistency.

Suggestions

- Strengthen test coverage. The two tests (valid email passes / leading NBSP fails) cover the reported bug but leave gaps worth closing directly rather than relying on the external test suite:
  - An explicit positive assertion that "joe bloggs"@example.com (ASCII space in a quoted local part) still passes — this is the main regression risk of the change, so guard it locally.
  - A second non-ASCII whitespace character (e.g. U+3000 IDEOGRAPHIC SPACE or U+2003 EM SPACE) to prove the fix generalizes beyond U+00A0.
  - Optionally trailing/interior NBSP placements.
- idn-email duplication. As the PR notes, IdnEmailFormat shares the identical delegation and therefore the identical bug (a NBSP is invalid in idn-email too — idn-email allows non-ASCII letters, not non-ASCII whitespace). Leaving it out of scope is defensible, but when the follow-up lands, containsNonAsciiWhitespace will be copy-pasted across the two near-identical classes. Consider extracting it to a shared static helper now so the two formats can't diverge later.

Broader observation (non-blocking)

The true root cause is that the bundled EmailValidator is permissive about non-ASCII input generally, not just whitespace. The email format is ASCII-only per RFC 5321, so one could argue any non-ASCII character should be rejected. The PR's narrow, whitespace-only scope is the right call for a targeted bugfix — it fixes #1164 without risking behavioral changes to real-world emails — but it's worth a sentence in the code/issue noting the fix is deliberately narrow.

Performance / Security

- Negligible cost: one short-circuiting O(n) code-point scan over an email-length string. No allocation of concern.
- No security risk; if anything this tightens input validation. No ReDoS surface (plain code-point predicate, no regex).

Verdict

Approve — correct, minimal, well-documented, and follows project test conventions. The only requested changes are optional polish: add the explicit quoted-local-part positive test and a second whitespace character, align the license header, and consider extracting the helper ahead of the idn-email follow-up.

PR #1272, opened 2026-08-15
Honor PathType when formatting SchemaException locations

### LINKED ISSUE #1270: SchemaException error location formatting
Greetings -
I'm using the `2.X` version of the library, and I've noticed that the new exception in `SchemaRegistry#validateSchemaNodeType` (and probably `#validateSchemaNodeNotNull` too) from https://github.com/networknt/json-schema-validator/issues/1174 / https://github.com/networknt/json-schema-validator/pull/1250 backported in `2.0.2` don't appear to be honoring the `SchemaRegistryConfig` settings I have in place for error location formatting.

My `SchemaRegistryConfig` is created like this, and is configured to use `PathType.JSON_PATH` for error locations:
```
private static final SchemaRegistryConfig SCHEMA_REGISTRY_CONFIG = SchemaRegistryConfig.builder()
    .pathType(PathType.JSON_PATH)
    .errorMessageKeyword("message")
    .cacheRefs(false)
    .formatAssertionsEnabled(true)
    .failFast(false)
    .preloadSchema(false)
    .typeLoose(false)
    .build();
```

If I feed in an intentionally invalid schema and validate it against the specification or dialect schema to check syntax and structure, I receive an error location like `$.properties.required` in my test case. But when I try and construct the invalid schema itself through the `SchemaRegistry`, I trigger the new `SchemaException` with a message of `Schema at #/properties/required must be object or boolean but was ARRAY` where the location of the same field is reported as `#/properties/required` instead. It seems to be using the default of `PathType.JSON_POINTER`.

Is this the expected behavior?

This new exception is also just using the string `message` instead of the `Error` object like other `SchemaException` instances with an error location I was already catching and handling. That's not a big deal and I've fixed that assumption in my code to fall back to the `message` if needed, but I was expecting the errors to have consistent formatting when referring to invalid elements in the schema across the various operations and methods.

kalayciburak:
## Summary
`SchemaRegistry` now formats `SchemaException` locations with the configured `PathType`. Creating a schema with `PathType.JSON_PATH` reports `$.properties.required` instead of always using the JSON Pointer `#/properties/required`.

Default `JSON_POINTER` messages are unchanged.

Fixes #1270

## Test Plan
- [x] `mvn -Dtest=Issue1270Test,Issue1174Test,SchemaLocationTest test` (Java 17)
- [x] `mvn test` (Java 17)

### DIFF
--- a/src/main/java/com/networknt/schema/SchemaLocation.java
+++ b/src/main/java/com/networknt/schema/SchemaLocation.java
@@ -392,6 +392,39 @@ public class SchemaLocation {
 
     }
 
+    /**
+     * Formats this location using the requested path type for the fragment.
+     * <p>
+     * Schema locations are stored as JSON Pointer fragments. Validation messages
+     * can request a different representation through {@link SchemaRegistryConfig}.
+     *
+     * @param pathType the path type to use for the fragment
+     * @return the formatted location
+     */
+    public String toString(PathType pathType) {
+        if (pathType == null || this.fragment == null || pathType == this.fragment.getPathType()) {
+            return toString();
+        }
+        NodePath converted = new NodePath(pathType);
+        int count = this.fragment.getNameCount();
+        for (int i = 0; i < count; i++) {
+            Object element = this.fragment.getElement(i);
+            if (element instanceof Number) {
+                converted = converted.append(((Number) element).intValue());
+            } else if (element != null) {
+                converted = converted.append(element.toString());
+            }
+        }
+        String fragmentText = converted.toString();
+        if (this.absoluteIri == null) {
+            return fragmentText;
+        }
+        if (fragmentText.startsWith("$")) {
+            return this.absoluteIri + fragmentText;
+        }
+        return this.absoluteIri + "#" + fragmentText;
+    }
+
     @Override
     public String toString() {
         if (this.value == null) {
--- a/src/main/java/com/networknt/schema/SchemaLocation.java
+++ b/src/main/java/com/networknt/schema/SchemaLocation.java
@@ -405,17 +405,11 @@ public class SchemaLocation {
         if (pathType == null || this.fragment == null || pathType == this.fragment.getPathType()) {
             return toString();
         }
-        NodePath converted = new NodePath(pathType);
-        int count = this.fragment.getNameCount();
-        for (int i = 0; i < count; i++) {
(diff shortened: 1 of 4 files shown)


stevehu:
The normal tests pass, but the new conversion corrupts anchor-based schema locations and can replace useful schema diagnostics when a property name is empty.

  Full review comments:

  - [P2] Preserve URI-reference anchor fragments — /home/steve/workspace/json-schema-validator/src/main/java/com/networknt/schema/SchemaLocation.java:405-405
    When the location contains an anchor, such as https://example.com/schema#anchor, its fragment has type URI_REFERENCE, but this condition allows it to be converted to the configured
    JSON_PATH. An invalid anchored schema consequently reports https://example.com/schema$.anchor, changing the anchor's identity; only JSON Pointer fragments should be converted, while
    anchors should remain #anchor.

  - [P2] Handle empty property names during JSONPath conversion — /home/steve/workspace/json-schema-validator/src/main/java/com/networknt/schema/SchemaLocation.java:418-418
    With PathType.JSON_PATH, an invalid subschema beneath an empty property name, such as {"properties":{"":[]}}, reaches this call with an empty path segment. PathType.JSON_PATH rejects that
    segment while rendering, so the intended must be object or boolean diagnostic is replaced by a wrapped IllegalArgumentException: A JSONPath selector cannot be empty; formatting must
    support or safely escape empty property names.

kalayciburak:
yep, anchors stay as #anchor now. empty names go to $[''] so the must be object or boolean msg isnt swallowed

stevehu:
The new conversion path cannot format schema locations containing empty object-property names and replaces the useful schema validation message with an internal formatting error.

  Review comment:

  - [P2] Handle empty keys when converting schema locations — /home/steve/workspace/json-schema-validator/src/main/java/com/networknt/schema/SchemaLocation.java:418-418
    When JSON_PATH is configured and an invalid subschema is beneath a valid empty property name, such as {"properties":{"":[]}}, converted.toString() throws because PathType.JSON_PATH
    rejects empty selectors. The resulting exception message becomes IllegalArgumentException: A JSONPath selector cannot be empty, masking the intended schema-type diagnostic; empty keys
    should be represented as $[''] or handled without invoking the rejecting formatter.

kalayciburak:
yep that empty key case is in the last commit already, it goes to $['']

PR #1274, opened 2026-08-17
Reject empty labels in idn-hostname format

dngr2:
The `idn-hostname` format accepts host names with a leading or interior empty label:

```
".example"   // reported valid — should be invalid
"a..b"       // reported valid — should be invalid
```

`RFC5892.isValid` skips empty labels with `continue`, with a comment about a trailing `.`. But `String.split` already drops trailing empty strings, so the skip only ever hits a **leading or interior** empty label, both of which are invalid. Rejecting them matches the JSON-Schema-Test-Suite (`idn-hostname.json`: "leading dot" and "empty label between two dots is invalid") and the reference validator. Ordinary host names (and trailing dots) are unaffected; the full suite stays green (added a test).

### DIFF
--- a/src/main/java/com/networknt/schema/utils/RFC5892.java
+++ b/src/main/java/com/networknt/schema/utils/RFC5892.java
@@ -99,11 +99,11 @@ public class RFC5892 {
         if ("".equals(value)) {
             return false; // empty string should fail
         }
-        if (value.length() == 1 && value.matches(LABEL_SEPARATOR_REGEX)) {
-            return false; // single label separator should fail
-        }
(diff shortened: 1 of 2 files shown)


dngr2:
Good catch. A value that is only separators (`...`, or `。。`) splits to an empty array, so the loop was skipped and it fell through to valid. Pushed a guard that rejects when there are no labels — that also subsumes the earlier single-separator case, so I dropped that special check. Added the two inputs as regression tests alongside the existing empty-label ones.

stevehu:
@copilot resolve the merge conflicts in this pull request

PR #1279, opened 2026-08-27
Fix nullable allOf ref type check

### LINKED ISSUE #1278: A property that is `nullable: true` and composed via `allOf` referencing a named schema does not accept null value
The following test should pass, since a null value is permitted by the `nullable: true`:

```java
@Test
void nullableAllOfRef() {
    String schemaData = """
            {
              "type": "object",
              "required": ["value"],
              "properties": {
                "value": {
                  "allOf": [ { "$ref": "#/components/schemas/Money" } ],
                  "nullable": true
                }
              },
              "components": {
                "schemas": {
                  "Money": {
                    "type": "object",
                    "required": ["amount"],
                    "properties": {
                      "amount": { "type": "integer" }
                    }
                  }
                }
              }
            }
            """;
    SchemaRegistry factory = SchemaRegistry.withDialect(Dialects.getOpenApi30());
    Schema schema = factory.getSchema(schemaData);

    List<Error> messages = schema.validate("{ \"value\": null }", InputFormat.JSON);
    assertEquals(0, messages.size());
}
```

tomakehurst:
Fixes issue where a nullable (OpenAPI dialect) on an element containing `allOf` isn't honoured when a null value is passed.

Closes #1278

### DIFF
--- a/src/main/java/com/networknt/schema/utils/JsonNodeTypes.java
+++ b/src/main/java/com/networknt/schema/utils/JsonNodeTypes.java
@@ -1,5 +1,7 @@
 package com.networknt.schema.utils;
 
+import java.util.Iterator;
+
 import tools.jackson.databind.JsonNode;
 import com.networknt.schema.ExecutionContext;
 import com.networknt.schema.Schema;
@@ -36,12 +38,9 @@ public class JsonNodeTypes {
             }
 
             if (nodeType == JsonType.NULL) {
-                if (parentSchema != null && schemaContext.isNullableKeywordEnabled()) {
-                    Schema grandParentSchema = parentSchema.getParentSchema();
-                    if (grandParentSchema != null && JsonNodeTypes.isNodeNullable(grandParentSchema.getSchemaNode())
-                            || JsonNodeTypes.isNodeNullable(parentSchema.getSchemaNode())) {
-                        return true;
-                    }
+                if (parentSchema != null && schemaContext.isNullableKeywordEnabled()
+                        && isNullableAncestor(parentSchema, executionContext)) {
+                    return true;
                 }
             }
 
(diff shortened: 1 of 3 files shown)


stevehu:
## PR review: Fix nullable allOf ref type check

  Risk: HIGH — request changes

  ### Finding

  1. [High] The change incorrectly honors nullable beside a direct $ref.

     In isNullableAncestor
     (https://github.com/networknt/json-schema-validator/blob/4307fda29e647f23c02a83bb28d5632b0ea28026/src/main/java/com/networknt/schema/utils/JsonNodeTypes.java#L95-L126), traversal returns
     the referencing $ref schema and then checks that schema’s own nullable property. Consequently, this now accepts null:

     {
       "$ref": "#/components/schemas/Money",
       "nullable": true
     }

     OpenAPI 3.0 requires additional properties on a Reference Object to be ignored, including nullable. The validator already implements that rule by discarding $ref siblings in
     Schema.read(). [OpenAPI 3.0.3 Reference Object specification](https://spec.openapis.org/oas/v3.0.3.html#reference-object) (https://spec.openapis.org/oas/v3.0.3.html#reference-object)

     I reproduced the behavioral regression:
      - Base a9c9638: correctly reports one type error.
      - PR head 4307fda: incorrectly reports zero errors.
      - The intended nullable + allOf + $ref case still passes.

     The traversal should inspect the lexical parent of a $ref wrapper—the allOf-owning Schema Object—without honoring nullable on the Reference Object itself. Add a regression test proving
     direct $ref siblings remain ignored.

  ### Verification

  - mvn -q test: passed on PR head.
  - Added focused cached-reference and chained-$ref probes: passed.
  - Direct $ref sibling regression probe: failed as described.
  - The linked [Copilot review](https://github.com/networknt/json-schema-validator/pull/1279#pullrequestreview-5044064268) (https://github.com/networknt/json-schema-validator/pull/1279#pullrequestreview-5044064268) generated no comments and missed this behavior change.

  Recommendation: REQUEST CHANGES before merge.

stevehu:
## Re-review: PR #1279

  No blocking findings. Risk: MEDIUM. Recommendation: APPROVE.

  The new commit 5d3d506 resolves the previous regression correctly:

  - Direct "$ref" + "nullable": true now remains rejected because the Reference Object’s own nullable is skipped.
  - "nullable": true on the allOf-owning schema still accepts null.
  - The contributor added a focused regression test that also asserts the resulting error is specifically a type error.
  - OpenAPI 3.1 remains unaffected.

  Verification on the current head:

  - Full mvn -q test: passed.
  - Cached target used from nullable and non-nullable reference sites: passed.
  - Multiple chained $ref schemas: passed.
  - Direct $ref sibling through a chained reference: correctly rejected.
  - OpenAPI 3.1 null union and ignored legacy nullable: passed.
  - GitNexus reports medium blast radius because JsonNodeTypes is shared validation infrastructure, but the changed branch remains gated by the nullable-enabled dialect.

  One non-blocking cleanup note: OpenApi30Test.java was normalized from CRLF to LF, making the diff look much larger than its functional change. That does not affect correctness.

  GitHub currently reports the [PR](https://github.com/networknt/json-schema-validator/pull/1279) (https://github.com/networknt/json-schema-validator/pull/1279) as mergeable, although no remote status checks are listed.

tomakehurst:
Thanks!

tomakehurst:
Should I open a 2.x PR also?

We're still on v2, so it'd be great to have the fix there too.

stevehu:
Yes. Please.

PR #1280, opened 2026-08-28
Port fix of nullable allof ref type check to the 2.x branch

tomakehurst:


stevehu:
No ReportFindings tool is available in this environment (I searched the deferred tool list — only GitNexus/Monitor/WebFetch etc. are exposed), so the findings are stated in full here.

What I did

- Resolved the target: networknt/json-schema-validator PR #1280, fix/nullable-allof-ref-type-check-2.x → base 2.x (a port of the master fix 8716b5b / PR #1279). Diff: JsonNodeTypes.java (+72) plus two test files.
- Built worktrees of both the PR head and origin/2.x, compiled both, and ran targeted differential probes plus the full test suite to confirm behaviour changes empirically.
- Full suite on the PR head: Tests run: 8424, Failures: 0, Errors: 74 — all 74 errors are GraalJS/ECMAScript regex NoClassDefFoundError from the local JDK 25 environment, unrelated to the diff. No existing test guards the regression below.

Findings

1. src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:107 — HIGH. nullable: true on a container schema now leaks into every $ref-ed child (properties / items / additionalProperties), accepting null where 2.x rejects it.

After findReferencingSchema hops to the $ref holder, line 107 unconditionally checks that holder's lexical parent for nullable. The javadoc assumes that parent is "the schema actually composing it (e.g. the allOf owner)", but the holder's lexical parent is identical in shape for allOf: [{$ref}] and for properties: {x: {$ref}} / items: {$ref} / additionalProperties: {$ref} — the code cannot tell them apart, so the parent's nullable is applied to the referenced value instead of to the container. Verified differential behaviour (OpenAPI 3.0 dialect, base 2.x vs PR head):

- {"order": {"nullable": true, "type":"object", "required":["customer"], "properties": {"customer": {"$ref": "#/components/schemas/Customer"}}}} with input {"order":{"customer":null}} → base: /order/customer: null found, object expected; PR: 0 errors. customer is not nullable.
- {"list": {"nullable": true, "type":"array", "items": {"$ref": ".../Money"}}} with {"list":[null]} → base: error; PR: 0 errors. This nullable array-of-$ref shape is extremely common in generated OpenAPI specs.
- Same for additionalProperties: {$ref} ({"m":{"k":null}}), and it also propagates through multi-hop $ref chains.

Net effect: null values silently validate against non-nullable referenced components. (For fairness: the inline equivalent — items: {"type":"object"} under a nullable array — already leaked on 2.x via the old grandparent check, so this is a widening of a pre-existing quirk rather than a brand-new class; but for $ref children it is a real regression against the base branch, and $ref is the dominant OpenAPI style.) Fix direction: ExecutionContext.getEvaluationSchemaPath() is maintained in lockstep with evaluationSchema (see AbstractKeywordValidator:133-153), so the walk can require that the keyword linking parent→holder is a composition keyword (allOf/anyOf/oneOf/not) before honouring the parent's nullable.

2. src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:110 — LOW. The walk stops at the first frame without $ref, so the fix misses the very pattern it targets one level deeper.

findReferencingSchema returns null unless the immediately preceding frame's node has $ref, so the walk cannot cross an intermediate composition frame. Both of these still fail on the PR head (same as base):
- value: {nullable: true, allOf: [{$ref: Money}]} where Money: {"allOf":[{"type":"object"}]} (a component that itself extends via allOf) → {"value":null} still reports null found, object expected.
- value: {nullable: true, allOf: [{allOf: [{$ref: Money}]}]} → same.

Not a regression, but the heuristic is arbitrary in the "too narrow" direction while being over-broad in the direction of finding 1; a keyword-aware walk would address both.

3. src/test/java/com/networknt/schema/oas/OpenApi30Test.java:115 — LOW. The javadoc on the new nullableAllOfRef test describes the pre-fix bug in the present tense, contradicting what the test now asserts.

It states equalsToSchemaType "only consults the schema that owns the failing type keyword and that schema's own parent" and that "the null-friendly check never fires and validation incorrectly fails" — that is exactly the code the same PR deletes. A reader hitting this test later will conclude the implementation still has the one-hop limitation. It should describe the expected behaviour, not the old defect.

Line-summary

- src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:107 — HIGH: unconditional parent-nullable check after the $ref hop makes nullable: true on an object/array propagate to $ref-ed properties, items and additionalProperties; {"order":{"customer":null}} and {"list":[null]} now validate where 2.x rejected them.
- src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:110 — LOW: walk halts at the first non-$ref frame, so nullable + allOf: [{$ref}] still fails when the referenced component itself uses allOf, or when the $ref sits inside a nested allOf.
- src/test/java/com/networknt/schema/oas/OpenApi30Test.java:115 — LOW: test javadoc documents the now-removed buggy behaviour as current, contradicting the assertion in the test body.

tomakehurst:
This now properly covers both the cases your review has highlighted plus a few extra. I'll await the next review before opening another PR to port the additional fixes back to main.

stevehu:
Reviewed networknt/json-schema-validator PR #1280 (fetched refs/pull/1280/head, base 2.x-line at ae64ed6; real change scope is cb258af~1...0a6feab, touching JsonNodeTypes.java plus two test classes). I built and ran the branch, plus the same probes against the pre-PR baseline, to confirm each finding empirically. Full existing suite: Tests run: 8434, Failures: 0 (74 errors are pre-existing GraalJS NoClassDefFound env issues, identical on baseline).

Note: no ReportFindings tool is exposed in this environment (ToolSearch returns no match), so the findings are stated below.

Findings (3)

- /home/steve/workspace/json-schema-validator/src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:130 — high. isComposingKeyword's non-index branch (lastElement instanceof Number ? fragment.getParent() : fragment) can never legitimately fire: allOf/oneOf/anyOf are all array-valued, so a real composing branch always ends in an index. The branch therefore only matches a property or definition literally named allOf/oneOf/anyOf, and treats it as a composing site. Confirmed regression: schema {"type":"object","properties":{"outer":{"nullable":true,"type":"object","properties":{"allOf":{"$ref":"#/components/schemas/Money"}}}}} with instance {"outer":{"allOf":null}} returns 0 errors on this branch but correctly returns /outer/allOf: null found, object expected on the pre-PR baseline. The same schema without the $ref (inline {"type":"string"}) also wrongly passes. The PR's own nullableAdditionalPropertiesDoNotNotLeakedWhenPropertyNamedAllOf / nullableItemsNotLeakedWhenPropertyNamedAllOf tests only pass because they insert an extra items/additionalProperties level between the collision and the type owner, so they give false confidence that the collision is handled. Fix: drop the non-index branch (require lastElement instanceof Number) now that not is excluded.
- /home/steve/workspace/json-schema-validator/src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:128 — medium. isComposingKeyword derives the relationship from the static schema.getSchemaLocation(), but Schema's constructor rewrites schemaLocation via resolve(...) whenever the subschema declares the dialect's id keyword — which is plain id for the OpenAPI 3.0 dialect (OpenApi30.ID_KEYWORD). When a composing branch carries an id, its fragment no longer ends in allOf/<n>, the check returns false, and nullable on the composing parent is silently dropped. Confirmed regression: {"type":"object","properties":{"value":{"allOf":[{"id":"http://example.com/inner","type":"string"}],"nullable":true}}} with {"value":null} yields /value: null found, string expected on this branch but 0 errors on the pre-PR baseline. Using the dynamic executionContext.getEvaluationSchemaPath() / evaluation path (as the neighbouring isEnumObjectSchema already does) avoids the id-rewrite hazard; the same static-vs-dynamic mismatch also means a $ref resolved by Schema.getSubSchema (which sets the resolved schema's parentSchema to the document root, not its lexical container, while keeping the full fragment) can read nullable off the wrong schema.
- /home/steve/workspace/json-schema-validator/src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:29 — medium. Adding oneOf to COMPOSING_KEYWORDS makes every branch's type keyword accept null, so a nullable oneOf with more than one branch now fails with a confusing "more than one match" instead of validating. Confirmed: {"properties":{"value":{"oneOf":[{"$ref":".../A"},{"$ref":".../B"}],"nullable":true}}} with {"value":null} reports /value: must be valid to one and only one schema, but 2 are valid with indexes '0, 1' — i.e. null is still rejected for the common two-branch nullable-oneOf shape. The added nullableOneOfRef test uses a single branch and so masks this; either add a multi-branch oneOf test and handle null above the branch loop (short-circuit in OneOfValidator when the composing schema is nullable), or document that only single-branch oneOf is supported.

Not flagged (checked and found sound): the $ref hop via findReferencingSchema picks the most recent stack occurrence, which is correct under recursion (Node→child→Node probes returned the right answers and did not loop); the deliberate exclusion of not is genuinely exercised by nullableNotDoesNotPropagateThroughNot; getElement(-1) on a root fragment returns null rather than throwing, so the null guard at line 131 is safe.

tomakehurst:
Hi @stevehu please could you share the configuration of the review agent/skill you're using?

I feel like this would be a lot easier if I could run it myself after making changes.

stevehu:
@tomakehurst The above review is just built in from Claude CLI with /review command. As this library is just a standalone component. For my other projects, I am using light-agent from light-fabric to drive codex and claude to mutually review each others work. The light-workflow is orchestrate all the activities. 

https://github.com/networknt/light-fabric

https://www.networknt.com/light-fabric/

stevehu:
Code review(medium · 3 findings)
src/main/java/com/networknt/schema/utils/JsonNodeTypes.java
  ●  94 [behavior-regression] Gating the parent hop on isComposingKeyword silently removes the long-standing propagation of `nullable` from a schema to its `properties`/`items`/`additionalProperties` subschemas, a behavior change on the 2.x maintenance branch that also diverges from the main-branch fix (#1279) this PR claims to port.
src/main/java/com/networknt/schema/keyword/OneOfValidator.java
  ● 201 [correctness]         `!nullableNode` suppresses the oneOf assertion for any null under a nullable ancestor, including when *no* branch matched, so branch constraints that legitimately reject null (enum, const, not) are silently dropped — and the same schema under anyOf still rejects.
  ●  69 [test-coverage]       The new comment states "Branches are still evaluated normally so their annotations and walk listeners still fire", but the existing `canShortCircuit` break after two valid branches still applies, so with three or more branches the remaining branches are never walked; the added test uses exactly two branches and cannot catch this.

  Ran 1 shell command

I reviewed PR #1280 (fix/nullable-allof-ref-type-check-2.x → 2.x) by building both the PR head and the 2.x base (ae64ed6) and diffing actual validation behavior. The full existing suite passes on the PR head (the only 74 errors are GraalJS/regex tests failing on Java 25, unrelated).

Findings

- src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:94 — Gating the parent hop on isComposingKeyword drops the old nullable propagation into properties/items/additionalProperties: {"outer":{"nullable":true,"type":"object","properties":{"name":{"type":"string"}}}} + {"outer":{"name":null}} is valid on 2.0.7 and now fails; main's merged #1279 keeps that check, so 2.x ends up stricter than the branch this "port" comes from.
- src/main/java/com/networknt/schema/keyword/OneOfValidator.java:201 — !nullableNode suppresses the oneOf error even when zero branches matched: {"nullable":true,"oneOf":[{"enum":["a"]},{"enum":["b"]}]} + null now yields 0 errors, while the same schema with anyOf still reports both enum errors. Narrowing to !(nullableNode && numberOfValidSchema > 1) keeps the intended fix.
- src/main/java/com/networknt/schema/keyword/OneOfValidator.java:69 — The comment promises branches still walk, but the pre-existing short-circuit after two valid branches still applies; with a third branch the walk listener fires twice, not three times. The new test uses exactly two branches, so it can't catch this.

Verified as non-issues along the way: no NPE path into OneOfValidator.validate with a Java-null node (the shouldValidateSchema && node != null guard matches AllOfValidator), no runaway loop or measurable cost in isNullableAncestor under deep/mutual $ref recursion (200 levels, ~2 ms), nullable does not leak into a $ref'd branch reused from a non-nullable site, and the id-rewrite test is meaningful since the OAS 3.0 dialect really does set idKeyword("id").

tomakehurst:
Separately from the review feedback, I did a deeper pass over this branch and found a cluster of problems in the ancestor-walk implementation that hadn't been raised. Rather than patch them one at a time, I rewrote the walk on top of information the engine already tracks.

## The change

`isNullableAncestor` reconstructed how a schema had been reached by climbing `getParentSchema()` pointers, identifying composing branches by `JsonNode` reference identity (`branch == schemaNode`), and inferring `$ref` hops by testing whether the previous stack frame's node happened to carry a `$ref` member.

None of that is necessary. `ExecutionContext` already maintains `evaluationSchema` and `evaluationSchemaPath` in lockstep — the schemas entered, and the keyword each one was entered by (pushed at `Schema.java:623`, and already consumed this way by `AbstractKeywordValidator.hasAdjacentKeywordInEvaluationPath`). Reading that makes the walk a single descent of a finite stack, with the hop keyword known exactly rather than guessed.

Net **−101 lines** in `src/main`: three heuristics collapse into one linear pass.

## Bugs this fixes

**Infinite loop on recursive schemas.** `findReferencingSchema` resolved a schema to its *topmost* occurrence on the evaluation stack, so when a schema appeared at two depths the walk could jump back up and oscillate. There was no visited set, depth cap or cycle guard.

```json
{"$ref": "#/$defs/X",
 "$defs": {"T": {"title": "t"},
           "X": {"$ref": "#/$defs/T", "type": "object",
                 "properties": {"c": {"$ref": "#/$defs/X"}}}}}
```

against `{"c": {"c": null}}`, on Draft 2020-12 plus a `nullable` `NonValidationKeyword`: base `ae64ed6` returns one `type` error in under a millisecond, this branch before the rewrite never returned at all — 100% CPU, no exception, no stack overflow. A second `allOf`-shaped reproducer behaved the same way. Both are now in `NullableAncestorTest` and fail by timeout against the pre-rewrite code.

Worth noting this needs a specific shape — a `$ref` target that also carries a `$ref` sibling plus a recursive property. A plain deep `$ref` chain doesn't trigger it, which is presumably why it went unnoticed.

**`enum` was never rewired.** The fix only ever applied to `type`. `EnumValidator` computed its accepted-value set in the constructor from `parentSchema`'s own node, so the headline case still failed whenever the referenced component was an enum:

```json
{"properties": {"value": {"allOf": [{"$ref": "#/components/schemas/E"}], "nullable": true}},
 "components": {"schemas": {"E": {"type": "string", "enum": ["a", "b"]}}}}
```

`{"value": null}` reported `enum @ /value`, while the same schema with the enum inlined beside `nullable` passed. Enum components behind `allOf`/`$ref` are the common OpenAPI shape, and `doc/1.x/config.md` states `nullable` affects both `enum` and `type`. The check now happens during validation, where the ancestor chain is known; existing error messages are unchanged.

This also turned out to fix the composing-keyword inconsistency raised in review at a deeper level than the suggested narrowing: because branches now accept `null` themselves, `oneOf`, `anyOf` and `allOf` agree on identical nullable schemas rather than returning 0/2/1 errors.

**Keyword coverage gaps.** `if`/`then`/`else` satisfy the "describes the same value" definition but were omitted, so `{"nullable": true, "if": …, "then": …, "else": …}` spuriously rejected `null`. `$dynamicRef` and `$recursiveRef` truncated the walk because only a literal `$ref` member was recognised. Both now have tests that fail against the pre-rewrite code.

**`$ref` into a composing branch.** A `$ref` such as `#/allOf/0` resolves to a schema whose parent is the document root, whose `allOf` array contains that very node — so the root's `nullable` was applied at an unrelated instance location. Using the recorded hop keyword closes this; also covered by a test that fails against the old code.

**Dialect boundaries.** `isNullableKeywordEnabled()` was consulted once, against the starting schema. With an unbounded walk that let a stray `"nullable": true` in a resource whose dialect has no such keyword suppress `type` assertions — plausible in hand-converted OAS 3.0 → 3.1 bundles. It is now re-checked at every hop.

**Correctness of the `not` exclusion.** The exclusion was documented but unenforceable: the old check only matched array-valued keywords, and `not` takes a single schema, so it could never have matched either way. It is now a real keyword check.

## Also in this branch

- `findReferencingSchema` rescanned the whole deque per hop, roughly O(depth²) per null value; now a single pass. Three lines of dead code removed alongside it.
- `isNullableAncestor` now takes only the `ExecutionContext`, so its dependence on live evaluation state is visible in the signature rather than an undocumented precondition on a `public` method in an exported package.
- `OpenApi30Test` repeated the same components block nine times and the same registry setup twenty-four times; those are shared now, and `allOf`/`oneOf`/`anyOf` are driven from one `@ParameterizedTest`.

## Verification

Full suite at each commit: **8453 tests, 0 failures**. The 74 errors are the pre-existing GraalJS/regex failures caused by the JDK blocking GraalVM's native access (JDK 26 locally), unrelated to this change and equally present on the base commit.

The behavioural claims above were checked by running the same probes against base `ae64ed6` and this branch side by side. Five of the seven cases in the new `NullableAncestorTest` fail against the pre-rewrite code — the two hangs, `if`/`then`/`else`, `$dynamicRef`, and the `$ref`-into-branch case.

Two items I addressed by construction but did not independently reproduce first, so flagging them as such: the `$ref`-member test misreading a sibling `properties` hop, and validation outcomes varying with `JsonNode` instance sharing. Both are eliminated by using the recorded hop keyword rather than inferring provenance, but I can't claim to have demonstrated either as a live bug.

stevehu:
Re-reviewed the 5 new commits (384b23b…b540a79) against a fresh build. Full suite green: 8396 tests, 0 failures, 0 errors (Graal/regex classes excluded — they fail on Java 25 for unrelated reasons).

All three earlier findings are resolved, and the fixes go further than patching symptoms:

- The ancestor walk was rewritten to descend the evaluationSchema / evaluationSchemaPath stacks instead of the lexical schema structure. The alignment it depends on holds — both stacks are pushed only in Schema.validate/walk, strictly nested, so sizes are equal whenever a keyword validator is executing, and the root schema correctly falls off the end. This also removes the identity-comparison heuristic and fixes the $ref-into-a-composing-branch false positive I'd noted as pre-existing (now covered by refPointingIntoComposingBranchDoesNotInheritNullable).
- oneOf now relaxes only the cardinality check (nullableNode && numberOfValidSchema > 1), and the short-circuit is disabled for a nullable null so all branches walk — the extended 3-branch test now genuinely fails against the old code. The companion EnumValidator change makes oneOf/anyOf/allOf agree on null under a nullable ancestor, which is what made the narrowing safe.
- The properties/items/additionalProperties behavior change is now documented in CHANGELOG with a migration note, and covered by tests.

Two remaining items:

- src/test/java/com/networknt/schema/OneOfValidatorTest.java:536 — nullableOneOfStillReportsErrorsWhenNoBranchMatches passes on the sibling not error; the single branch actually matches, so the fix it guards can be reverted with the test still green. {"oneOf":[{"not":{"type":"null"}}],"nullable":true} exercises the real path.
- src/main/java/com/networknt/schema/utils/JsonNodeTypes.java:135 — viaReference drops a nullable sibling of $ref in every dialect, but siblings are only ignorable pre-2019-09. On the 2020-12+nullable dialect the new test file introduces, one nullable: true is honoured for the holder's own type and ignored for the target's.

Neither blocks the OpenAPI 3.0 behavior the PR targets; the first is the one I'd fix before merge, since it's the regression guard for the change made in 392b127.

stevehu:
Re-reviewed at `d2741ab` (2 new commits since the last pass: `131f4be`, `d2741ab`). Built the head and base `ae64ed6` side by side, ran differential probes, the full suite, and mutation-tested each new guard.

**Full suite: 8454 tests, 0 failures.** The 74 errors are the pre-existing GraalJS/`NoClassDefFound` failures from the local JDK, identical on base.

## Both prior items are fixed — and the tests genuinely guard them

I mutation-tested rather than trusting the green suite, since "the test guards nothing" was the substance of the first item:

| Reverted change | Test that fails |
| --- | --- |
| `nullMatchedMoreThanOneBranch && …` → `!nullableNode` | `OneOfValidatorTest.nullableOneOfStillReportsErrorsWhenNoBranchMatches:547` |
| `!(viaReference && ignoresReferenceSiblings(schema))` → `!viaReference` | `NullableAncestorTest.nullableBesideRefIsHonouredWhenDialectKeepsSiblings` |
| `numberOfValidSchema > 1 && !nullableNode` → `numberOfValidSchema > 1` | `OneOfValidatorTest.nullableMultiBranchStillWalksBranches` |

`ignoresReferenceSiblings` mirrors `Schema.java:565` exactly — same expression, same threshold — so the walk and validator assembly can't drift apart.

## Two new findings, both low

**1. `CHANGELOG.md:14` — the `oneOf` cardinality relaxation isn't in the changelog.** The two bullets cover the ancestor walk (`type`, `enum`) and the container change, but not the `OneOfValidator` behaviour that `392b127` settled, which is the largest user-visible improvement here. `{"nullable":true,"oneOf":[{"$ref":".../A"},{"$ref":".../B"}]}` with `null` goes from 3 errors (`oneOf` + both branch `type`s) on 2.0.7 to valid. A reader upgrading from a "more than one match" error won't find it described. The `not` change is in the same position (`{"nullable":true,"not":{"type":"string"}}` + `null` goes invalid → valid) but it is at least pinned by `nullableNotDoesNotPropagateThroughNot`.

**2. `src/main/java/com/networknt/schema/keyword/EnumValidator.java:79` — `typeLoose` + `nullable` + `enum` changes for the string `"null"`.** Dropping `nodes.add(NullNode.getInstance())` also removed `NullNode` from the set `isTypeLooseContainsInEnum` scans, and that method compares `n.asText()`, which is `"null"` for a `NullNode`. Confirmed on the OAS 3.0 dialect with `typeLoose(true)`:

```
{"properties":{"v":{"enum":["a"],"nullable":true}}}  +  {"v":"null"}
  2.0.7 → valid        d2741ab → enum error
```

Head is the more defensible answer — the literal string `"null"` should not satisfy `enum: ["a"]` — but it's an unguarded behaviour change on a maintenance branch under a non-default config, so it's worth either a test or a changelog line.

## Checked and sound

Member order around `$ref`/`nullable` doesn't affect the result in either dialect; `$dynamicRef` and `$recursiveRef` siblings behave like `$ref`; a component shared between a nullable and a non-nullable site stays strict at the non-nullable one; `required`/`minimum`/`pattern`/`minLength` under a nullable `allOf`+`$ref` still skip `null` correctly and still fail a bad object; the recursive-schema case terminates; `enum` messages are byte-identical to 2.0.7; no public API was removed, so the branch stays binary compatible. Performance is unchanged — 20,000 validations of a 60-level nullable/`allOf` nest with a null leaf ran in 271 ms on head vs 280 ms on base.

Neither finding blocks the OpenAPI 3.0 behaviour the PR targets. I'd merge after a changelog line for the `oneOf` change.

stevehu:
I found no bugs in PR #1280, either in the latest commit (09deacf, which adds changelog entries and a typeLoose test) or in the nullable changes it builds on.

How I checked:
- The new stack walk: I read the isNullableAncestor walk in JsonNodeTypes. The schema stack and keyword stack are only pushed and popped together in Schema.validate and Schema.walk, so they stay in step. The keyword names it checks (allOf, anyOf, oneOf, if, $ref, $dynamicRef, $recursiveRef) are the names those validators actually put on the stack.
- $ref siblings: ignoring a nullable written next to $ref in pre-2019-09 dialects matches how Schema already drops those siblings.
- Base vs. PR: I built both the 2.x base and this PR and ran the same set of OpenAPI 3.0 cases against each. Every behaviour change matches what the new changelog entries say:
  - A nullable container no longer makes its properties, items or additionalProperties children accept null.
  - A nullable oneOf now accepts null, including with fail-fast on.
  - {"nullable": true, "not": {"type": "string"}} now accepts null.
  - With typeLoose on, the string "null" is now rejected by a nullable enum, while a real null is still accepted.
  - allOf + $ref, nested anyOf/allOf, and enum reached through $ref all now accept null.
  - nullable written next to $ref is still ignored, as before.
  - walk with validation gives the same results as validate.
- Full test suite: 8,455 tests ran with no failures. The 74 errors are all in the GraalJS regex tests, which can't load their regex engine on this machine; they have nothing to do with this PR.

PR #1281, opened 2026-09-15
Honor PathType when formatting DiscriminatorValidator schema locations

youdie006:
Follow-up to #1272. That routed the `"Schema at ..."` messages through `SchemaRegistry.formatSchemaLocation`, which renders the location with the configured `PathType`, and applied it to the two throws in that class. `DiscriminatorValidator` throws the same message at three sites and still concatenates the raw `SchemaLocation`.

With `pathType(PathType.JSON_PATH)` the two halves of one sentence disagree, because `instanceLocation` is already a `NodePath` built from the configured `PathType`:

```
Schema at #/allOf/1/discriminator is redefining the discriminator property that has already been set for at $
```

#1270 asked for "consistent formatting when referring to invalid elements in the schema across the various operations and methods". These three are the remaining `"Schema at " + SchemaLocation` sites in `src/main`.

<details>
<summary>Verification</summary>

Java 17, `mvn -B test` in `maven:3-eclipse-temurin-17`. Two tests added to `Issue1270Test`, one per direction.

| row | `DiscriminatorValidator.java` md5 | `Issue1270Test` |
|---|---|---|
| master | `f6675b68e0c64bce8eaeae4ae70f99df` | **1 failure**: the JSON_PATH row; #1272's own three rows pass |
| this change | `45882fae985972f68617437fbea3ecd6` | 5 of 5 pass |
| hard-code `PathType.JSON_PATH`, ignoring the config | `72c4c3d41b329a655a071ba81bc98f12` | **1 failure**: the default row, which now reads `$.allOf[1].discriminator` |

The two added tests fail on different mutants, so neither is redundant: one catches "does not honour the configured type", the other catches "ignores the configuration". The default `JSON_POINTER` output is unchanged.

Full suite on the change: `Tests run: 8504, Failures: 0, Errors: 0, Skipped: 30`, `BUILD SUCCESS`. That includes `Issue1270Test` and `SchemaLocationTest`, so #1272 is unaffected.

Reproducer, no fixtures needed: `{"allOf":[{"discriminator":{"propertyName":"a"}},{"discriminator":{"propertyName":"b"}}]}` validated against `{}` under `Dialects.getOpenApi31()`.

Site `:134` is the one the test drives. `:119` needs `strict("discriminator")` and `:157` needs a conflicting explicit mapping; both take the identical one-line change. Nothing in `src/main` or `src/test` parses these message strings, so no behaviour is pinned to the old text.

No `CHANGELOG.md` hunk, since the entries there are written at release time.

</details>

Disclosure: written with AI assistance (Claude). Every number above is measured.


stevehu:
I found no bugs in PR #1281.

The change swaps the raw this.schemaLocation for a new private formatSchemaLocation() helper in the three SchemaException messages in DiscriminatorValidator (lines 119, 134 and 157). The helper produces the same text as SchemaRegistry.formatSchemaLocation, which #1272 added:
- Null config: schemaContext.getSchemaRegistryConfig() is already called without a null check on line 118, so the helper doesn't add a new place where it can crash.
- Null path type: SchemaLocation.toString(PathType) treats a null path type as plain toString().
- Default output: the default is JSON_POINTER, which matches how the fragment is stored, so the old message text comes out unchanged.
- Anchors: locations that are anchors rather than JSON Pointers still print unconverted.
- Coverage: no other src/main exception message still glues a raw schemaLocation into its text. The only other concatenations are debug toString() methods.

One thing is left over from #1272 and this PR doesn't cause it: under JSON_PATH, a schema location with a base IRI prints the IRI directly followed by $…, with no # between them.

stevehu:
It isn't needed for correctness. It only changes the wording of an error message, in a situation most users will never hit.

What it fixes: Each of these errors names two locations in one sentence. The instance location already follows the configured PathType, but the schema location was always printed as a JSON Pointer. So with pathType(PathType.JSON_PATH) you'd get:

Schema at #/allOf/1/discriminator is redefining the discriminator property that has already been set for at $

After the change, both halves use the configured format ($.allOf[1].discriminator … at $). That closes out the consistency request in #1270; #1272 already made the same fix in SchemaRegistry.

How much it matters: very little.
- Few people see the error: it only fires with the OpenAPI dialect and conflicting discriminator keywords, which is a schema-authoring mistake.
- Fewer still see a change: the text only differs for users who set a non-default PathType. With the default (JSON_POINTER) the message is exactly the same.
- No behaviour changes: it's still the same exception at the same point, and nothing in the codebase reads these message strings.

Downsides:
- Duplicated helper: it adds a private formatSchemaLocation() that repeats the one in SchemaRegistry. That's harmless, though a shared helper would be tidier.
- Base IRI formatting: when the schema has a base IRI, JSON_PATH output joins it straight onto the path with no # (for example https://x/schema.json$.allOf[1].discriminator). That's arguably less readable than the old https://x/schema.json#/allOf/1/discriminator. The behaviour comes from #1272's SchemaLocation.toString(PathType), but this PR puts it into three more messages.

Verdict: it's safe, well tested and consistent with #1272, so it's reasonable to merge as a small cleanup, but it isn't urgent. If the base-IRI format bothers you, fix it in SchemaLocation.toString(PathType) before or together with this PR, rather than rejecting the PR.

PR #1283, opened 2026-09-16
Port the nullable ancestor walk from the 2.x branch

tomakehurst:
Resolves the OpenAPI nullable keyword by descending the engine's own evaluation stacks rather than reconstructing provenance from schema structure. The previous walk climbed lexical parents and sniffed for $ref members, which misattributed nullable across reference sites, was sensitive to schema-node identity, and missed $dynamicRef and $recursiveRef entirely.

The walk pairs evaluationSchema with evaluationSchemaPath to recover the keyword each schema was entered by, propagating nullable through allOf/oneOf/anyOf, if/then/else and the three referencing keywords, and stopping at any keyword that moves to a different value. As a result nullable no longer leaks into a container's properties, items or additionalProperties children, and no longer reaches inside not.

Applies the same walk to enum, which previously decided at construction time whether null was accepted, and relaxes only the oneOf cardinality check for a nullable null, leaving branch errors reported when nothing matches.

Ported from networknt/json-schema-validator#1280. Two adaptations were needed for this branch rather than a straight copy:

  - isNodeNullable now calls asBoolean(false). Jackson 3 throws on a node it cannot coerce where Jackson 2 returns false, and the walk reads nullable on every composing ancestor, so a malformed member anywhere in the chain would abort the validation.

  - The typeLoose enum change concerns the empty string here, not the string "null". NullNode.asString() is "" in Jackson 3, so that is the value the accepted set silently permitted.


stevehu:
I found no bugs in PR #1283.

What I checked:
- The two stacks line up. Schema.validate and Schema.walk push one schema, then push and pop one keyword name per validator. Branch indexes go on a different path. So skipping the top keyword in isNullableAncestor does give the keyword each schema was entered by, the same way hasAdjacentKeywordInEvaluationPath already reads it.
- Keyword names match what gets pushed. KeywordType.IF_THEN_ELSE is "if", which is the name IfValidator runs under, and $ref/$dynamicRef/$recursiveRef/allOf/anyOf/oneOf also match.
- ignoresReferenceSiblings matches how schemas are built. In Schema.getValidators, a draft before 2019-09 keeps only the $ref validator and drops its siblings. OpenAPI 3.0 is DRAFT_4, so a nullable next to a reference is skipped. A nullable on a schema that composes that reference through allOf still counts.
- Every caller runs inside a validation. TypeValidator, including plain types in a type array, EnumValidator and OneOfValidator all run with the owning schema on top of the stack, so the walk starts at the right place.
- OneOfValidator only loosens the "exactly one match" rule when the value is a nullable null and more than one branch matched. When no branch matches, the branch errors are still reported. Turning off short-circuiting for that case has no effect on the result.
- Tests: NullableAncestorTest, OneOfValidatorTest and OpenApi30Test all pass on the PR branch (offline Maven run).

Some behaviour changes are deliberate and listed in the CHANGELOG, so I didn't count them as bugs:
- nullable no longer reaches into properties, items or not.
- With typeLoose, an enum no longer accepts the empty string.
- A nullable next to a root $ref is ignored.

I used a temporary git worktree for the PR and removed it afterwards.

stevehu:
Short answer: yes, it can break some existing apps, but only apps that depend on nullable behaving in ways the OpenAPI 3.0 spec doesn't allow. The change makes sense, and it is a faithful port of the 2.x change with two small edits for Jackson 3.

Is it just a port?

Mostly. I compared the finished JsonNodeTypes, EnumValidator and OneOfValidator on origin/2.x (after #1280 merged) with this PR's branch. The nullable logic is identical. The only differences are:
- Jackson 3 fixes. asBoolean(false) is used so a malformed nullable value can't throw mid-validation. The typeLoose note in the CHANGELOG says the empty string instead of "null", because Jackson 3 turns a JSON null into "" when read as text.
- An extra CHANGELOG entry. It says a nullable next to a root $ref is now ignored. The 2.x CHANGELOG leaves this out, but 2.x changed the same way.

The two diffs look different for another reason. Master already had #1279, an earlier partial fix, and 2.x did not. But #1279 was never released: no tag contains it. So measured against what users actually run (3.0.7 and 2.0.7), both branches go through the same behaviour change.

What changes for existing apps

Both released versions use the same simple rule: null passes a type check if the schema or its direct lexical parent says nullable: true. Moving from that rule to the new walk:

┌────────────────────────────────────────────────────────────────────────────┬────────────────────────┬───────────────────────────────────────────────────────────┐
│                                   Change                                   │         Effect         │                        Break risk                         │
├────────────────────────────────────────────────────────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────┤
│ nullable reaches through allOf/oneOf/anyOf/$ref at any depth               │ More null values pass  │ Low. Only schemas that declare nullable accept more nulls │
├────────────────────────────────────────────────────────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────┤
│ nullable on an object/array no longer covers its properties or items       │ Fewer null values pass │ Main risk                                                 │
├────────────────────────────────────────────────────────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────┤
│ Nullable oneOf no longer fails with "more than one branch matched" on null │ More null values pass  │ Low. The old result was a false rejection                 │
├────────────────────────────────────────────────────────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────┤
│ nullable no longer applies inside not                                      │ More null values pass  │ Low                                                       │
├────────────────────────────────────────────────────────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────┤
│ typeLoose + nullable enum stops accepting "" (3.x) or "null" (2.x)         │ Fewer values pass      │ Very low                                                  │
├────────────────────────────────────────────────────────────────────────────┼────────────────────────┼───────────────────────────────────────────────────────────┤
│ nullable next to a root $ref is ignored                                    │ null rejected          │ Low to medium                                             │
└────────────────────────────────────────────────────────────────────────────┴────────────────────────┴───────────────────────────────────────────────────────────┘

The real risk is the container case. Take this schema:
{"type": "object", "nullable": true, "properties": {"name": {"type": "string"}}}
Today {"name": null} passes, because the old rule looks one level up and finds nullable on the object. After this PR it fails. OpenAPI specs often end up relying on this by accident. A light-rest-4j service could start returning 400s on requests that currently pass, without any change to its spec.

Does it make sense?

Yes:
- In OpenAPI 3.0, nullable describes only the schema it sits on. A nullable object doesn't make its fields nullable.
- Siblings of a $ref are ignored.
- The old grandparent rule was right in some cases only by chance. It was wrong for $ref chains and for nesting deeper than one level, and it leaked into children.

The new walk follows the path validation actually took, which is the correct model.

Recommendation

Merge it, but treat it as a behaviour change, not a quiet bugfix:
1. Call out the container change prominently in the release notes, with the before/after example above. The CHANGELOG already says to add nullable on the child itself.
2. Before releasing, run the light-rest-4j / light-4j OpenAPI validator test suites against a snapshot. Those are the most likely places to hit it.
3. 2.x already merged the same change, so both lines stay consistent. Consider adding the root-$ref CHANGELOG entry to 2.x too.

If you'd rather not break anyone in a patch release, the other option is a registry config flag to keep the old container behaviour for one release. I wouldn't do that by default, since it keeps a behaviour that goes against the spec.

